PR #13, opened 2019-08-07
IASC-561 replace jquery dropdown with vanilla

left23:
**Note**: I will close this without merging. I will keep it open until I transfer the info below to some docs because it is a good description of the expected behaviour of the various interactive pieces in the common header.

Compare with https://ochabasic.dev.ahconu.org to see the Bootstrap dropdown and jquery version.
OCHA Services, search, the mobile menu toggle and main menu dropdowns, and the language switcher all use Bootstrap dropdown plugin.

### cd-dropdown
[cd-dropdown.js](https://github.com/UN-OCHA/common_design/pull/13/files#diff-c357641e3e7c7ad0799d8fc855066431) is to replace Bootstrap dropdown.js plugin, and so requires specific markup to function correctly.

OCHA services menu is a good example. [See cd-ocha.html.twig](https://github.com/UN-OCHA/common_design/pull/13/files#diff-16d269a3d1ebc25e2c6174ab7bda3b0e)

There are 3 elements.
1. The main wrapper element
2. The button element that toggles the dropdown
3. The element that drops down (and its contents)

The main wrapper receives a class of `js-open` when the button is clicked, and this class is used for styling the child elements.

The button needs `data-toggle="cd-dropdown"` and `aria-expanded="false"`. The JS uses the [data-toggle attribute](https://github.com/UN-OCHA/common_design/pull/13/files#diff-c357641e3e7c7ad0799d8fc855066431R5) to add the click event listener, and [toggles the aria-expanded value](https://github.com/UN-OCHA/common_design/pull/13/files#diff-c357641e3e7c7ad0799d8fc855066431R8)

The element that drops down should be [the sibling of the button](https://github.com/UN-OCHA/common_design/pull/13/files#diff-c357641e3e7c7ad0799d8fc855066431R15) and have `class="cd-dropdown"`. We use `.js-open .cd-dropdown` selector to [set the display value with CSS](https://github.com/UN-OCHA/common_design/pull/13/files#diff-dd11447ac1b1618d46aacc1599dd5eb8R8).

When [any area outside the dropdown is clicked, the dropdown is closed](https://github.com/UN-OCHA/common_design/pull/13/files#diff-c357641e3e7c7ad0799d8fc855066431R28).
TODO When another dropdown element is clicked, the currently active dropdown should close.

### cd-search
cd-search has [cd-dropdown as a dependency](https://github.com/UN-OCHA/common_design/pull/13/files#diff-8479bf6d461a726e01208373027f0cb1R17) 
As this is drupal specific, I use a preprocess function to add the [cd-dropdown class to the form](https://github.com/UN-OCHA/common_design/pull/13/files#diff-e4c3d4de0a2e7f3c93bb9e7d04c02cbcR45)
The button that toggles the dropdown is part of the [cd templates](https://github.com/UN-OCHA/common_design/pull/13/files#diff-526fb546c7b1166240ac58f84b3668f2R22)

In [cd-search.js](https://github.com/UN-OCHA/common_design/pull/13/files#diff-1a577310fc69fa02bba1a34d0231e33c) I force focus on the input when the button is toggled.
TODO: the submit button used to have a class added too, so its colour changes when the input is in focus. I'd like to do this in CSS with adjacent sibling selectors but this seems too markup dependent
eg. `.cd-search input:focus + button[type=submit] {}
In [Drupal's case](https://github.com/UN-OCHA/common_design/pull/13/files#diff-b7da0d624fbd5d976e1c2f86d9db9a79R193), it adds loads of wrapper divs around both input and button elements so JS seems to be the only way. I'm in denial about this so currently the submit button does not change colour.

### cd-menu
For the menu, we use the cd-dropdown for the mobile menu toggle and the main menu items that have dropdowns.
The mobile menu button markup is part of the [cd templates](https://github.com/UN-OCHA/common_design/pull/13/files#diff-901ab8bffb3b43e65ac0844bbff73ac0)
The class `cd-dropdown` is added to the [Drupal classes array](https://github.com/UN-OCHA/common_design/pull/13/files#diff-077fa83820c51c32175522402f77797bR41) in the main menu block template override.
The main menu items that have dropdowns have the [necessary dropdown-related attributes](https://github.com/UN-OCHA/common_design/pull/13/files#diff-621eb56f43233dab30dbf570bb6757f1R49) in menu--main.html.twig
In cd-menu.js (), `.menu-item--expanded a` as a function to prevent the click event on the anchor.

The display of the mobile menu and the main menu dropdown menus are controlled in CSS, first with the cd-dropdown display rule used for all dropdowns but also set in [cd-nav.scss](https://github.com/UN-OCHA/common_design/pull/13/files#diff-49464b9fa4eac86a21f0059baccf1fc6) to reinforce the proper display when viewport widths change between media queries. 

### cd-language-switcher TODO
I have not looked at this in D8 at all yet


rupl:
Is this the PR with IE11 compat issues? The only thing that stuck out to me was `let` which can often work, but IIRC not with `forEach` loops.

If that's the only ES we want to take advantage of, we might as well stick with `var` since transpilers just convert them back to `var` anyway.

left23:
@rupl Yes, this is it. Thanks
Here's what I added to get around the IE issue
https://github.com/UN-OCHA/common_design/pull/13/commits/bccee6a36d04dfb3a041f5ed1b41667b7a790abe
More to be discussed :)

rupl:
ahhh gotcha. yeah, bummer but in this case I think LoE is _waaaay_ lower to rewrite a few lines. Also, if we introduce a transpiler, we either have to save artifacts or require others to use the same toolchain. I think in this case it's just simpler to go with old JS syntax.

PR #15, opened 2019-09-30
Corrects declaration of dependencies in common_design.libraries.yml

berliner:


left23:
Thanks @berliner Well spotted

PR #18, opened 2019-11-26
IASC-561 improvements to form alter hook for Views exposed forms

left23:
Based on the use case of IASC
https://github.com/UN-OCHA/iasc8/pull/203/files#diff-f56012826bd853edaca2440bb162e6a1
I've left the empty array in the base theme, and a comment explaining how it should be used in the subtheme.

This should allow easier switching between the various search form implementations I've seen in implementation.

I've also adjusted the form placeholder operator which would throw an error when there was a placeholder entered in the Views exposed form via the GUI


left23:
@attiks Do you think this is a reasonable way to deal with the Views exposed form alter in the base theme?
https://github.com/UN-OCHA/common_design/pull/18/files#diff-e4c3d4de0a2e7f3c93bb9e7d04c02cbcR56

The subtheme.theme would then need the hook with the Views ID, if I wanted to affect a Views exposed form:
`
function subtheme_common_design_form_alter(&$form, FormStateInterface $form_state, $form_id) {

  $includeView = [
    'search-content-search',
    'views-exposed-form-search-content-search'
  ];

  // If in array above, add search--inline attributes.
  if (in_array($form['#id'], $includeView)) {

    // This is for a Views exposed form search block.
    if ($form_id == 'views_exposed_form') {
      $form['#attributes']['class'][] = 'cd-search__form';
      $form['#attributes']['aria-labelledby'][] = 'cd-search_btn';
      $form['keys']['#attributes']['placeholder'] = t('What are you looking for?');
      $form['keys']['#attributes']['class'][] = 'cd-search__input';
      $form['keys']['#attributes']['type'][] = 'search';
      $form['keys']['#attributes']['id'][] = 'cd-search';
      // Theme suggestion for submit element.
      $form['actions']['submit']['#attributes']['data-twig-suggestion'] = 'search_submit';
      $form['actions']['submit']['#attributes']['class'][] = 'cd-search__submit';
      $form['actions']['submit']['#attributes']['value'][] = 'Search';
    }
  }
}
`

common_design.theme
```diff
@@ -65,7 +64,7 @@ function common_design_form_alter(&$form, FormStateInterface $form_state, $form_
     if ($form_id == 'views_exposed_form') {
```
attiks: You don't need this line anymore, only the in_array one

left23: +1

PR #20, opened 2019-12-23
CD-177 improvements

left23:
This PR integrates the RW javascript for the dropdowns, using custom data attributes and creating the toggle buttons with Javascript. This is nice for progressive enhancement because if JS is not enabled, we don't have stray useless buttons in the markup.
I've added a few elements to the RW script to accommodate what already exists in the common header and footer, like the dropdown arrow icon and the ocha logo for the Services menu.
Additionally, the main menu needs special treatment when it has nested menus that also expand on click.

### To do:

 - Refactor [cd-dropdown](https://github.com/UN-OCHA/common_design/blob/894b33ee8fb98ac25a4557799216c88420f354cd/js/cd-dropdown.js) and [cd-menu.js](https://github.com/UN-OCHA/common_design/blob/894b33ee8fb98ac25a4557799216c88420f354cd/js/cd-menu.js) so functions are reusable and menu-related code can remain in cd-menu. Also consider the selector for the menu may change so provide a way to add selectors in the subtheme.

- Styling for when JS in disabled (needs improvement)

- Polyfills for IE. See if we can load a JS file based on browser detection (the lte IE 8 conditionals were dropped as of IE10) TBD


left23:
@rupl Over to you. If it's nicer for you, feel free to merge this and open a fresh PR.

rupl:
Ok sounds good, then I'll merge it and start fresh from the main branch.

PR #21, opened 2020-01-08
📌 [OPS-6125] Move required deps from devDeps, so they are present for production builds.

cafuego:
If they remain in devDepencencies, they will not be installed when the production container is built with `NODE_ENV=production` and cause a build failure.

See https://github.com/UN-OCHA/common-design-site/commit/ed18c886e0f47463a7c4eb61d1a3898d234211e7/checks?check_suite_id=389465190

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "common-design",
-  "version": "2.0.0",
+  "version": "2.0.1",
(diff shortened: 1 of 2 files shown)


cafuego:
No worries. I made a 2.0.1 release so that I can get a working container for the common-design-site.

PR #22, opened 2020-01-13
CD-183: reusable Drupal behaviors

rupl:
# CD-183

Jiggling around some functions to make them reusable across different files. Renamed all event handlers to follow a convention and threw some docs in to show others (or ourselves in 8 months) how the files should be structured.

As I understand D8 Libraries, we should now be able to use each one as a dependency and it will bundle everything up in the correct order, allowing us to pick and choose which files get loaded. They're all pretty small but if any one file grows significantly, this can help us cut back on unused code being shipped with the theme.

_Edit:_ [whitespace-free diff](https://github.com/UN-OCHA/common_design/pull/22/files?utf8=%E2%9C%93&diff=split&w=1) to help see what actually changed


js/cd-menu.js
```diff
@@ -1,20 +1,8 @@
...
-        for (var i = 0, l = elements.length; i < l; i++) {
-          var element = elements[i];
-          setToggable(element, element.previousElementSibling);
-        }
-      }
-
-        // Update nested Drupal menus in the header.
-        updateDrupalTogglableMenus();
+      // Update nested Drupal menus in the header.
+      Drupal.behaviors.cdDropdown.updateDrupalTogglableMenus('.cd-header .cd-ocha-btn + .cd-dropdown');
```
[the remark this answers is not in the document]
left23: The cd-menu JS is for the main menu so the selector should probably match the D8 drupal for the main menu (I can set this up on the demo site). However, it's a good test to add my own selectors so I'm happy to test it this way first.

rupl:
ok I think I cleaned this up after resolving the conflicts from the linting PR. @left23 up to you whether you want to test again, or we merge and react to any problems I created 😉

js/cd-dropdown.js
```diff
@@ -3,297 +3,302 @@
...
+     * Create a button to toggle a dropdown.
+     */
+    createButton: function (element) {
+      var label = element.getAttribute('data-toggable');
+      var logo = element.getAttribute('data-logo');
+      var icon = element.getAttribute('data-icon');
+      var component = element.getAttribute('data-component');
+
+      // Create the button.
+      var button = this.context.createElement('button');
```
orakili: @left23 @rupl  Not sure if `context` is ever anything other than the `document` but if that's the case and it's a DOM element for example, then the `createElement` function will not exist and an error will be thrown. (Same with the createElementNS in the createIcon function etc.)

rupl: `context` can be something other than `document`, for example when Drupal Ajax is invoked. [Here's a d.o comment explaining it](https://www.drupal.org/forum/support/module-development-and-code-questions/2013-02-19/what-is-context-for-in-drupal#comment-13074787).

However for our custom Behavior, we have reasonable certainty that it will always be `document`, and I suppose we could just simplify.

PR #25, opened 2020-01-20
CD-187 components

left23:
See [CD-187](https://humanitarian.atlassian.net/browse/CD-187) for details

[Components](https://github.com/UN-OCHA/common_design/tree/cd-187-components/components) currently live in the theme's root. 

The components inform the markup, styles, layout behaviour and interactivity, if any, of the elements that may make up a typical web page.

There is no way to view them in the browser without some assembly.
Example: https://commondesign.demo.ahconu.org/themes/contrib/common_design/components/cd-teaser/cd-teaser.html is missing the CSS because this is not being served as a Drupal page.

Some (cd-hero and cd-teaser) also has a Drupal implementation for demo purposes and for drupal implmentations.

For Drupal, we can use twig files and the [Components](https://www.drupal.org/project/components) module to define [component libraries](https://github.com/UN-OCHA/common_design/blob/cd-187-components/common_design.info.yml#L37).

This means we can have Drupal templates in the theme's root, and include them using the namespace `@cd-components`
Eg. [page--front.html.twig](https://github.com/UN-OCHA/common_design/blob/cd-187-components/templates/page--front.html.twig#L58) can include the [cd-hero component](https://github.com/UN-OCHA/common_design/tree/cd-187-components/components/cd-hero).

We also define the[ components as libraries](https://github.com/UN-OCHA/common_design/blob/cd-187-components/common_design.libraries.yml#L11) and attach them in [Drupal's twig templates](https://github.com/UN-OCHA/common_design/blob/cd-187-components/components/cd-hero/cd-hero.html.twig#L1) in a modular fashion.

### cd-alert
cd-alert
### cd-article-teaser
cd-article-teaser
### cd-bullet-list
cd-bullet-list
### cd-list
cd-list
### cd-teaser
cd-teaser
### cd-user-list
cd-user-list

PR #26, opened 2020-01-22
CD-83 refactor main menu

left23:
This is branching off #25 so shouldn't be merged until after.

I've refactored the main menu rules to use the new selectors as a result of #22 cd-dropdown refactoring.
I've added button element selectors where needed. When a Drupal menu item is set via the GUI to be expanded, the element will render as a button instead of an anchor. This will be especially handy when the JS is disabled because we can hide the button elements.

I've simplified where possible, removing any 3rd level menu styling.

I made a few changes to z-index when I noticed the cd-hero component in the content area overlapped the main menu dropdown. I've added a z-index variable for the main menu for clarity.


sass/cd/cd-header/_cd-nav.scss
```diff
@@ -75,6 +70,8 @@
   float: left;
 
   .cd-nav {
+    @include z-index(main-menu); // zindex may need adjusting depending on other site content.
```
left23: @rupl FYI You might notice the main menu dropdowns do not appear because of a z-index issue
 https://commondesign.demo.ahconu.org
Until this PR is merged, you can either remove the cd-hero include in page--front.html.twig
https://github.com/UN-OCHA/common_design/pull/25/files#diff-841e2d37dcc55695d28dffc8974728c6R58
or add `z-index: 199` for .cd-nav

PR #30, opened 2020-01-29
CD-98 user menu

left23:
See https://humanitarian.atlassian.net/browse/CD-98


left23:
Merging this to then deploy for collab on demo

PR #34, opened 2020-02-04
CD-186 JS polyfills

left23:
- Conditionally loading for browsers that do not support es6 modules
- Set as dependency for cd-dropdown
- Remove unneeded polyfills and testing-related code

### DIFF
--- a/common_design.libraries.yml
+++ b/common_design.libraries.yml
@@ -7,6 +7,8 @@ global-styling:
 cd-dropdown:
   js:
     js/cd-dropdown.js: {}
+  dependencies:
+  - common_design/cd-polyfill
 
 cd-menu:
   js:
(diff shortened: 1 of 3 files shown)


common_design.libraries.yml
```diff
@@ -86,8 +88,8 @@ cd-user-list:
       components/cd-user-list/cd-user-list.css: {}
 
 
-
-#This doesn't work for IE11 support, due to conditionals being dropped in IE10
+# The nomodule attribute means browsers who support ES6 modules will ignore this file but browsers like IE load it.
+# Also deferring the file so it loads after everything else.
```
rupl: ah, nice find!

One thing: I typically load polyfills before any other code, so that the functionality they provide can be available to any custom JS we write.

rupl: Time for me to "well actually" myself: I double checked [MDN's `<script>` article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) and it does seem ok to use `defer` since it will wait until the DOM is parsed, but execute *before* `DOMContentLoaded` fires. So if any code that uses polyfilled features is wrapped in a dom-ready listener, it will be ok.

Again, good find on the `nomodule` attr, and ignore me while I refresh my understanding of these various props. 😆

PR #52, opened 2020-04-07
CD-187 add sidebar second layout

left23:
Classes are added to the `main` wrapper element [when blocks are rendered](https://github.com/UN-OCHA/common_design/pull/52/files#diff-5d225d2e3e8538dcc0aa6138eae8454aR49) in `sidebar-first` or `sidebar-second` region.

eg.
```
<main cd-container cd-layout-with-sidebar-first>
  <aside cd-layout-sidebar-first>
  <div cd-layout-content>
```
or
```
<main cd-container cd-layout-with-sidebar-both>
  <aside cd-layout-sidebar-first>
  <div cd-layout-content>
  <aside cd-layout-sidebar-second>
```

### Changes of note:

1. I added a sidebar-second Drupal region.

2. I swapped the source order. We used to only have sidebar first, and it was on the right. Now we have sidebar-first on the left (also in source order) and sidebar-second (new) on the right. This means site's using the base theme with see their sidebars on the left now.

3. I refactored cd-layout partial to include a second sidebar, added vertical rhythm for all viewports, simplified selectors to ease overriding rules, kept all permutations as separate instances eg. sidebar-both, sidebar-left, sidebar-right so to allow for as much customisation as possible.

4. I changes the conditional printing of the wrapper divs around the sidebar region, removing the filters in favour of using the CSS :empty pseudo class to hide divs that render when the region renders empty. This is to do with an ongoing Drupal issue and the implementation here has caveats. I added comments: https://github.com/UN-OCHA/common_design/pull/52/commits/3e8b9411b37e98891937be7301108600bc81828d#diff-5d225d2e3e8538dcc0aa6138eae8454aR74

### Expected result
Desktop with both sidebars
Screenshot from 2020-04-07 18-09-34
Tablet with both
Screenshot from 2020-04-07 18-09-11

Mobile with both
Screenshot from 2020-04-07 18-08-51

Sidebar First
Screenshot from 2020-04-07 18-10-06

Sidebar Second
Screenshot from 2020-04-07 18-10-32


templates/page.html.twig
```diff
@@ -71,7 +71,7 @@
     <main role="main" {{ attributes.addClass(layout_classes) }}>
       <a id="main-content" tabindex="-1"></a>{# link is in html.html.twig #}
 
-      {% if page.sidebar_first|render %}
+      {% if page.sidebar_first|render|striptags|trim is not empty %}
         <aside class="cd-layout-sidebar-first" role="complementary">
```
left23, line 75: Ensuring the sidebars are rendering when desired is proving difficult.
See https://www.drupal.org/node/953034 for history. This is an ongoing Drupal issue.
When there are blocks in regions that are access controlled (eg. Tools block, displays for admins when logged in), the empty region div with print for anon users.
When I add filters `|striptags|trim is not empty ` to `page.sidebar_second|render` the Facets in the sidebars don't render (i.e. no sidebar prints). The only results where I currently get Facets printing in sidebar (for either anon and auth) is to use `page.sidebar_second|render` only

PR #55, opened 2020-04-13
Duplicate toggle button in edge case scenario

berliner:
It's all a bit complicated but I try to explain to the best of my abilities.

All mentioned methods are to be found in [cd-dropdown.js](https://github.com/UN-OCHA/common_design/blob/master/js/cd-dropdown.js) if not otherwise indicated.

I have run into a problem in FTS Public on pages with a toggable header menu ("Help", as can be seen here for example: https://dev.fts8.dev.ahconu.org/appeals/overview/2019) and also display facet filters as checkboxes. On those pages, I observe 2 toggle buttons, one coming from the twig template file (see below) and the second one get's created by `initializeToggables`.

The base situation:
We have added the Help menu to the common header following our understanding of how this should be done and basically copying [menu--help.html.twig](https://github.com/UN-OCHA/common_design/blob/master/templates/navigation/menu--help.html.twig) as `menu--utility.html.twig` into our subtheme, same for [block--system-menu-block--help.html.twig](https://github.com/UN-OCHA/common_design/blob/master/templates/block/block--system-menu-block--help.html.twig) that got copied into the subtheme as `block--system-menu-block--utility.html.twig` (different names because our menu is called utility internally).

The problem(s):
The main problem I see is that `initializeToggables` works on the context that is passed in to `Drupal.attachBehaviors` while `updateDrupalTogglableMenus` works on the full document.
Another problem that I see is that `initializeToggables` doesn't take the possibility into account, that their might already be a button, like is the case for the help menu added to the header.
The facet module that we use on some pages calls `Drupal.attachBehaviors` with a restricted context before it gets called the first time with the full document context.
Another potential problem form my understanding is that `[data-cd-toggable]` is used both to identify elements that should receive a toggle button, but it is also added to the nested menu tree by the `setToggable` method ("so that it can be handled properly by the global click handler"), resulting in potentially double-usage and misidentification.

The flow for the problem I observe is roughly like this for the help menu:
1. `Drupal.attachBehaviors(RESTRCITED_CONTEXT)` gets called
2. `updateDrupalTogglableMenus` gets called with arguments from `Drupal.behaviors.cdUserMenu` and finds elements matching the selector against the full document, but also passes `element.previousElementSibling`, which is a button, to the `setToggable` method, therefor the button gets it's attributes and event handlers and the base element get's the `[data-cd-toggable]` attribute which will create problems in step 4, but the button itself not added again as it is already there
3. `Drupal.attachBehaviors(FULL_DOCUMENT)` gets called
4. `initializeToggables`gets called, and finds elements to work on based on the `[data-cd-toggable]` selector which already has been added to the menu list item by `setToggable` in step 2, then calling `setToggable` again but without the `toggler` argument, thus adding a new button for the menu, resulting in a duplicate button on our help menu.

I am able to fix this problem in 2 different ways:
1. Applying the change that this PR is about, so that `updateDrupalTogglableMenus` always works on the context that is setup when `Drupal.behaviors.cdDropdown.attach` is called.
2. Modifying `initializeToggables` so that it also passes in `element.previousElementSibling` to the `setToggable`method. In case there is no previous sibling this will be `null` and shouldn't interfere, in case there is a sibling, it will prevent a second button to be created.

Sorry, this got long, but I'm actually still a bit confused about how the parts play together.


berliner:
I found another issue that is slightly complicated to explain too. In general I really see quite some issues with the current code in `cd-dropdown.js`, which makes it particularly unsatisfying to debug, maybe we can discuss this some time?

The issue at hand here relates to the problem that Hardik already mentioned in https://humanitarian.atlassian.net/browse/HPC-6495 about the `initializeToggables` function not finding the main menu for logged in users, which results in the the mobile menu button not being created. The reason for that is that FTS Public D8 uses Drupal's [BigPipe rendering technique](https://www.drupal.org/docs/8/core/modules/big-pipe/overview) for logged-in users (not sure if it came installed by default or if we enabled it on purpose, but it's quite good so I don't see a reason for not using it). At the time when `initializeToggables` is first called, the part of the DOM that later becomes the main header navigation looks like this:
```
<span data-big-pipe-placeholder-id="callback=Drupal%5Cblock%5CBlockViewBuilder%3A%3AlazyBuilder&amp;args%5B0%5D=mainnavigation&amp;args%5B1%5D=full&amp;args%5B2%5D&amp;token=nTQyzF-mrHLZ6vZIFWhf-QQEsK10jNLcAELsI76f7io"></span>
```
This is the reason that Hardik saw only a single element returned by the call to `this.context.querySelectorAll`.

Now BigPipe is nicely integrated with Drupal 8 and so it doesn't miss to correctly call `attachBehaviors` for each part of the DOM that it loads after the initial page load, only that when it comes to the main menu, the context variable looks like this:
```
<nav id="cd-site-header__nav" role="navigation" class="contextual-region block block-menu navigation menu--main cd-nav" aria-labelledby="cd-nav-toggle" data-cd-toggable="Menu" data-cd-component="cd-nav" data-cd-hidden="true" data-cd-icon="arrow-down">
  ...
</nav>
```
A call to `.querySelectorAll('[data-cd-toggable]')` on that context variable doesn't find anything, because there is no **child** element of `<nav>` that has the `data-cd-toggable` attribute.

So, my additional change to this PR basically does this to fix the problem.
Whenever `initializeToggables` gets called, it not only looks for child elements of context that have the `data-cd-toggable` attribute, but it also checks whether the context itself has that attribute, in which case the context variable itself gets added to the list of found elements that are passed on to `setToggable`. This fixes the problem for me and I can now see the mobile menu button for both logged-in and anonymous users.

left23:
@berliner
The issue you describe above with Bigpipe should be solved in the latest release.
https://github.com/UN-OCHA/common_design/releases/tag/v2.8.0
Feel free to test at your leisure and let me know if you encounter issues, or have suggestions and/or open PR directly.

PR #61, opened 2020-05-04
CD-79 CD-111 Better typography defaults

left23:
This PR includes a nice way to manage fonts in a base theme

From the various implementation requirements to date (like webfonts, or conflicting font rules) setting the font family in the header and footer only, is how we ensure consistency in those elements without  affecting the content area.

[DSR](https://github.com/UN-OCHA/reports-site/blob/2a6f209d9dcce7e398dbb60c09383105d29e16df/layouts/default.vue#L103) has a nice implementation for font family in the content area, based on our discusson on CD-111 and https://markdotto.com/2018/02/07/github-system-fonts

I think this is a lovely default so I've added the system font family to the body, as part of the common_design base theme. We can override it when specific implementations need it.

I've also added an example for using webfonts, comment out, as a helper.

This will affect some existing implementations, but in a good way. The implementations that already have overrides in place (very possible) will not change.

before/after screenshot
Selection_005


left23:
@rupl Feel free to input, no need to test locally. It's mostly from the DSR repo :)

PR #76, opened 2020-06-23
CD-201 dropdown conditions

left23:
1. refactor cd-dropdown.js to accommodate a condition based on a media query and a CSS custom property `--dropdown`, which denotes when the toggler (button) should be set, unset and updated. 
2. bind event handler pattern applied to all event handlers
3. add resize event handler in intializeToggables function, and move main loop from same to handleResize function
4. Comment out some CSS that sets display rules which should now be moot because button is destroyed/recreated. Also helpful for debugging.
5. Set default CSS custom property in cd-dropdown partial on `data-cd-toggable`
```
[data-cd-toggable] {
  --dropdown: true;
}
```
6.  set CSS custom property on cd-nav element, to use in cd-dropdown.js to detect when we want the dropdown to be destroyed/recreated. In this case, on desktop, we don't want the main menu hidden by a toggle. On mobile, we do. This is an example of how we can use this on other dropdown elements, like the collapsible filters on mobile.
7. use document instead of context to find all elements with attribute `data-cd-toggable`, to prevent issues for auth users when bigpipe is enabled, as described in https://github.com/UN-OCHA/common_design/pull/55#issuecomment-615415546. In the past we had issues using document (resulting in multiple buttons being created, described in https://github.com/UN-OCHA/common_design/pull/55#issue-402696857) but this should be corrected because we have added a [condition to check if a button already exists](https://github.com/UN-OCHA/common_design/pull/76/commits/bdd2fac946ac6795a23035100614dff659e3e248#diff-c357641e3e7c7ad0799d8fc855066431R253)

Background discussion on [CD-201](https://humanitarian.atlassian.net/browse/CD-201)


left23:
@orakili Thanks. This all works very well. The additional comments are helpful. I think this is a nice improvement. If you want to tweak further, feel free to do so directly. I'll leave this PR open for a few days.
@rupl Casual review if you are interested :)

I haven't cross-browser tested yet. I expect failures around the lack of support for CSS custom properties. I'm using a ponyfill for IE11 elsewhere that might be appropriate.

I might also have some more tidying up display-related rules that are no longer needed so I will be committing more on this branch.

PR #78, opened 2020-07-06
[CD-201] dropdown conditions - additions

orakili:
Additional changes to the cd-dropdown.js file.

### DIFF
--- a/js/cd-dropdown.js
+++ b/js/cd-dropdown.js
@@ -313,7 +313,7 @@
       // Hide the element.
       element.setAttribute('data-cd-hidden', expand === false);
 
-      // Add the toggler before the toggable element id not already.
+      // Add the toggler before the toggable element if not already.
       if (element.previousElementSibling !== toggler) {
         element.parentNode.insertBefore(toggler, element);
       }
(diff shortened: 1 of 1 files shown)


js/cd-dropdown.js
```diff
@@ -257,21 +257,28 @@
...
-      // Skip if the toggler has already been created.
-      if (toggler && toggler.hasAttribute('data-cd-toggler')) {
-        return;
+      // Skip if the toggler is not a button or has already been processed.
+      if (toggler) {
+        // Togglers should be buttons to avoid mis-processing elements
+        // appearing before the toggable element. There is still a risk of
+        // mis-processint if, for whatever reason, there is a button which is
+        // not the toggler before the toggable element.
+        if (toggler.nodeName !== 'BUTTON' || ) {
```
left23, line 266: The hanging `||` -  should there be an OR condition? Or should this be removed?
Removing it makes everything work for me

orakili: Sorry, I split the 2 conditions to better explain their purpose and missed that... Thanks for spotting the problem.

Issue #87, opened 2020-07-15
Event listener is not attached to button when it is recreated by cd-dropdown JS


left23:
https://github.com/UN-OCHA/common_design/blob/48cb59cda351f063a12cefb8ecc834a3dc7aade1/js/cd-search--inline.js#L14

When the search is toggled on mobile, the input should be given focus (it is styled with a drop shadow). This still works on initial page load but when the viewport is resized from desktop to mobile, and the search toggler button is removed and recreated, the event listener is not attached so input is not given focus when the Button is toggled.

orakili:
@left23 Untested, but maybe the adding the event on the parent could do the trick:

```javascript
(function (Drupal) {
  'use strict';

  Drupal.behaviors.cdSearchInline = {
    attach: function (context, settings) {
      var searchInput = context.querySelector('.cd-search--inline__input');

      function focusInput(event) {
        var searchButton = context.querySelector('.cd-search--inline__btn');
        if (searchButton && event.currentTarget === searchButton) {
          searchInput.focus();
        }
      }

      if (searchInput) {
        searchInput.parentNode.addEventListener('click', focusInput);
      }
    }
  };
})(Drupal);
```

left23:
@orakili Would you mind putting this in a PR if you have a solution? Thanks!

left23:
Thanks @orakili 
I'll put this into a PR soon and I think it means I can remove the cd-search.js and cd-search--inline.js files too which is great!

PR #93, opened 2020-07-30
Issue-87 search input focus

left23:
We can use `data-cd-focus-target` attribute to focus the input when toggler is clicked as explained in #87 

This means I can remove the 2 custom js files whose only purpose was to deal with focus on the input.

This also means I can remove the 2 library definitions and I no longer need to attach those libraries to the search templates.
However, the libraries included [cd-dropdown.js as a dependency](https://github.com/UN-OCHA/common_design/commit/e49f703366d4c2d0a04e5a950ff956e9dd21bc38#diff-8479bf6d461a726e01208373027f0cb1L27) and since I am removing these libraries, this will no longer be the case.

cd-dropdown is needed for the searches behaviour. At the moment, it is being attached to the [cd-navigation template](https://github.com/UN-OCHA/common_design/blob/master/templates/cd/cd-header/cd-navigation.html.twig#L1), as [a dependency of cd-menu.js](https://github.com/UN-OCHA/common_design/blob/master/common_design.libraries.yml#L12
) which means it is loaded on all pages and thus available for the search too. It is pretty likely that the cd-navigation template will be present in most implementations but maybe there are some scenarios where this won't be the case.

Do people find this clear? Is there something I can do better to communicate that the search's need the cd-dropdown for behaviour? Or should I consider including it as part of the global library?


orakili:
@left23 Given that the cd-dropdown.js is used for the OCHA services menu which is present on all the pages, I think it would be kind of logical to have it as a global library.

PR #97, opened 2020-07-31
[CD-212] Change the skip to content link target to be more in accordance with latest recommendations

orakili:
Ticket: https://humanitarian.atlassian.net/browse/CD-212

Currently the target of the skip to content link is an `<a>` tag with the `main-content` id. This is an old method and not the recommended one, which is to add the id directly to the main sectioning element (`main` in our case): https://webaim.org/techniques/skipnav/

Drupal has an [open issue](https://www.drupal.org/project/drupal/issues/2784311) to patch core that's taking forever to be added because of the impact on bartik and other themes. That's however not an issue for us so we can change the `page.html.twig` template in the base common design theme.

### DIFF
--- a/templates/maintenance-page.html.twig
+++ b/templates/maintenance-page.html.twig
@@ -40,8 +40,8 @@
 
   </header>
 
-  <main role="main" class="cd-container">
-    <a id="main-content" tabindex="-1"></a>{# link is in html.html.twig #}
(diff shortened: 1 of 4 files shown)


left23:
@lazysoundsystem @orakili This change might need to be reflected in implementation sub themes if the page templates have been overridden.

PR #102, opened 2020-08-19
CD-issue 99 & 100 Error page templates

left23:
Replace text string with `page.content` region so generic error message appears on a 404 page *I assume the text string was there to provide a friendlier message and remains a nice option for specific implementations, but can be added to the sub theme instead.

Add class `page-error` and `page-error--404` to body class for 403 and 404 status codes, to facilitate styling of these pages.


left23:
Adding reviewers for FYI

PR #105, opened 2020-08-19
CD-206 better SVG icon default sizes

left23:
See discussion on [CD-206](https://humanitarian.atlassian.net/browse/CD-206) & commits


left23:
Adding reviewers for FYI. Feel free to comment here or on JIRA issue.

PR #107, opened 2020-10-13
CD-222 remove decorative SVGs from accessibility tree

left23:
Add aria-hidden true and focusable false to icon element to remove decorative SVGs from accessibility tree and prevent issues reported in firefox testing. Screenshots of issue on [issue](https://humanitarian.atlassian.net/browse/CD-222).
Also add colour for footer menu h2 to correct another contrast-related issue in FF.
Results should be no Accessibility issues reported in FF.


rupl (approved):
Nice improvement!

lazysoundsystem (approved):
Makes sense

orakili (approved):
Works well. 

I think the https://github.com/UN-OCHA/common_design/blob/master/README.md should be updated as well:

```
<svg class="cd-icon cd-icon--arrow-down" width="16" height="16">
  <use xlink:href="#cd-icon--arrow-down"></use>
</svg>
```

To

```
<svg class="cd-icon cd-icon--arrow-down" width="16" height="16" aria-hidden="true" focusable="false">
  <use xlink:href="#cd-icon--arrow-down"></use>
</svg>
```

left23:
Thanks @orakili I've adjusted the README and added a note about removing the icons from the accessibility tree.

PR #108, opened 2020-10-14
CD-221 broken aria-labelledby and missing menuitem role

left23:
As per [CD-221](https://humanitarian.atlassian.net/browse/CD-221), I've correct the errors from WAVE tool.

For the search and language switcher, I added an ID to the block h2, and its value as aria-labelledby value.

For the OCHA Services and also language switcher block, I've added menuitem as role to the li elements. However I could use a second opinion on this. It's unclear to me whether the role=menu should be used, and if not, the menuitem role is not relevant.

### DIFF
--- a/common_design.theme
+++ b/common_design.theme
@@ -53,7 +53,7 @@ function common_design_form_alter(&$form, FormStateInterface $form_state, $form_
   // This is for Drupal core search block.
   if ($form_id == 'search_block_form') {
     $form['#attributes']['class'][] = 'cd-search__form';
-    $form['#attributes']['aria-labelledby'][] = 'cd-search-btn';
+    $form['#attributes']['aria-labelledby'][] = 'cd-search-form';
     $form['#attributes']['data-cd-toggable'][] = 'Search';
     $form['#attributes']['data-cd-icon'][] = '';
     $form['#attributes']['data-cd-component'][] = 'cd-search';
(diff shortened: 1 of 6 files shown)


orakili (changes requested):
The changes regarding the labels and titles look good.

However, the use of the `menuitem` role should be avoided according to https://inclusive-components.design/menus-menu-buttons/ (see the "THE MENU AND MENUITEM ROLES" section):

> Since our navigation menu is beginning to behave somewhat like a “true” menu, should these not be present?

> The short answer is: no. The long answer is: no, because our list items contain links and [menuitem elements are not intended to have interactive descendants](https://w3c.github.io/html-aria/#index-aria-menuitem). That is, they are the controls in a menu.

> We could, of course, suppress the list semantics of the `<li>`s using role="presentation" or role="none" (which are equivalent) and place the menuitem role on each link. However, this would suppress the implicit link role. In other words, the example to follow would be announced as “Home, menu item”, not “Home, link” or “Home, menu item, link”. ARIA roles simply override HTML roles.

The last paragraph above matches the example from the w3.org site: https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html but Heydon argues that's not a good idea:

> We want the user to know that they are using a link and can expect link behavior, so this is no good. Like I said, true menus are for (JavaScript driven) application behavior.

So it looks like in any case the `role="menuitem"` shouldn't be used on the `li`.

I would remove the "menu" and "menuitem" roles altogether.

left23:
Thanks for confirming @orakili 
I've made those changes.

orakili (approved):
Looks good, thanks!

PR #110, opened 2020-10-29
CD-226 Design System

left23:
Feature branch for Common Design v3
This includes design and code improvements and refactoring to common_design base theme, and Components for the design system.


left23:
@lazysoundsystem @rupl @orakili I've added a [CONTRIBUTING.md](https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/CONTRIBUTING.md) file which links to the OPS ticket that explains how the repo (base theme and common-design-site) are connected and how they auto-deploy.
Also, I've update some dependencies including replacing `node-sass` with `dart-sass` so you should run `npm install` to update.

left23:
> There are now two variants for banners.

If you are referring to adding the `cd-bleed` and `cd-bleed--background-only`, I don't consider adding additional components to be variants (there might be lack of clarity with vocabulary here). 
Unlike the `cd-hero` where the variation is specific to the core component Eg. `cd-hero--style-one` means nothing without `cd-hero` styles, whereas the `cd-bleed` is more like a utility because it is not banner-specific (similar to `cd-caption`) and can be used with other components.

left23:
> Very picky, but could this be 24rem?

Yep, this height came from the designs but increasing it by 14px isn't going to make a difference (says the developer :smile: )

left23:
> Now we have some variants, the caption text could describe them.

I'm doing my best to not add description text into the Demo page to describe the components, in the short term. I don't want the maintenance burden and it might lead to discrepancies between the READMEs and the Demo page, which I want to avoid.
We should treat the READMEs as the authority, and the demo page as relatively disposable. If/when the time comes where we think more about packaging this like a component "catalogue" (meant for others outside our team), we can devote time in coming up with good ways to display and explain.

components/README.md
```diff
@@ -0,0 +1,136 @@
+### Component QA status
+
+Add a check if you've reviewed the component and it passes the Criteria.
+
+| Component           | QA round 1          | QA round 2          |
```
left23, line 5: Add  semantic titles to replace "QA round 1, 2"

components/cd-alert/cd-alert.css
```diff
@@ -1,31 +1,90 @@
...
   margin: 1rem 0 2rem;
+  border-radius: 0.25rem;
+  border: 1px solid var(--cd-blue--bright);
+  box-shadow: -8px 0 0 var(--cd-blue--bright);
+  color: var(--cd-grey--dark);
+}
+
+.cd-alert,
+/* This is Drupal specific. If it's an error message, it gets an extra div with a role. */
+.cd-alert div[role='alert'] {
```
rupl, line 13: I wonder if we should just include the wrapper unconditionally and set the `role` based on the type of alert.

It would allow us to set one rule for flex container, and avoid implementation errors in the non-drupal websites where markup is also being replicated (e.g. for HID I didn't know this is the case and our system can't even cater to this at present)

components/cd-alert/cd-alert.css
```diff
@@ -1,31 +1,90 @@
...
+}
+
+.cd-alert--error {
+  background: var(--cd-red--light);
+  border: 1px solid var(--cd-red);
+  box-shadow: -8px 0 0 var(--cd-red);
+}
+
+.cd-alert--error svg {
+  fill: var(--cd-red);
```
rupl, line 50: The pattern here (and the others on the parent such as `background`, `border-color` and `box-shadow` color seem like repetition of selectors/properties that CSS Custom Props are supposed to do away with.

When I plan custom props in a component, I think of setting component-specific "arguments" then filling those args with values:

```css
.cd-alert {
  /* define component defaults */
  --cd-alert-color-primary: var(--color-1);
  --cd-alert-color-background: var(--color-2);

  /* use defaults */
  background: var(--cd-alert-color-background);
  border: 1px solid var(--cd-alert-color-primary);
  box-shadow: -8px 0 0 var(--cd-alert-color-primary);
}

.cd-alert svg {
  /* use defaults */
  fill: var(--cd-alert-color-primary);
}

/**
 * Now inside our variants, we ONLY set custom props and NEVER set any other CSS.
 * It's mostly like a function in PHP/JS now, with nothing being set except args.
 * 
 * This ensures that we don't fat-finger the box-shadow into being 9px or other
 * mistakes like that. Additionally, we don't have to repeat selectors for SVG.
 */
.cd-alert--error {
  --cd-alert-color-primary: var(--cd-red);
  --cd-alert-color-background: var(--cd-red--light);
}

File: components/cd-utilities/cd-utilities.css [hunk 4]
```diff
@@ -0,0 +1,66 @@
...
+
+/* Rounded image */
+/* Requires square image */
+*[class*="__image--rounded"] {
+  display: block;
+  border-radius: 50%;
+  overflow: hidden;
+  /* For IE11 */
+  width: 100%;
+}
```
rupl, line 48: I found my way here by way of the `cd-byline` and its rounded image.

I wonder if that should be made slightly more generic? just `--rounded` should do, or perhaps if it were its own standalone class: `.cd-util--rounded` — then the class name denotes where it came from, and isn't tied to a particular sub-element (in this case a class containing `__image`).

Most of these rules seem more like global defaults rather than utilities to me. If I've missed out on important reasons why they're here, please bop me on the head 😄  If not, maybe we can discuss on the next call and see what we could do to either make them more generic and utility-like.

components/cd-bleed/cd-bleed.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Update the --gho-bleed-scrollbar-width css variable when the <html> element
+ * is resized, including when the scrollbar is initially added.
+ */
+(function () {
```
rupl, line 5: I prefer to name all functions, since it allows stack traces to better identify where errors arise. I typically call the top-level IIFE `iife()` — it doesn't matter if the name is reused in multiple files since it's wrapped in parentheses.

components/cd-bullet-list/README.md
```diff
@@ -0,0 +1,18 @@
...
+None
+
+### Variants
+The default bullet list uses the SVG element data URI as a background image. This means the fill colour can still be 
+edited in CSS.
+
+The other variant `--icon` is when the SVG element markup is added to the list.
+
+```
+cd-bullet-list--icon
```
rupl, line 16: the other READMEs I saw prefixed variants with dots to indicate that it's a CSS class

components/cd-bullet-list/cd-bullet-list.html.twig
```diff
@@ -1,31 +1,51 @@
...
+      <svg class="cd-icon cd-icon--arrow-right" aria-hidden="true" focusable="false" width="16" height="16">
         <use xlink:href="#cd-icon--arrow-right"></use>
       </svg>
-      Menu item title
+      Adipiscing egestas mauris est viverra nibh
+    </a>
+  </li>
+</ul>
+
+<ul class="cd-bullet-list [ cd-flow ]">
```
rupl, line 33: I'm not sure if this is out of scope for the individual components, but the demo displays these two lists as one. I can't see where one ends and another begins.

components/cd-button/cd-button.css
```diff
@@ -41,42 +42,51 @@
 }
 
 /* Default button */
-.cd-button {
+.cd-button,
+input[type='submit'] {
```
rupl, line 46: I wonder if we should be applying this to stuff that isn't explicitly marked with `cd-button`. These selectors for `input` turn the component into more of a CSS reset, but its presence depends on one instance of the `cd-button` component being output by Drupal.

In its current form, it would apply styles unevenly if the page has `<input type="reset">` for example. I don't know if Drupal uses that markup, but HID does. No need to overhaul based on one edge case; It would be far easier to change HID. But I still wanted to highlight the possibility.

components/cd-button/cd-button.css
```diff
@@ -153,57 +196,98 @@ a.cd-button {
...
-  }
+.cd-button--icon:hover svg,
+.cd-button--icon:focus svg {
+  fill: var(--cd-white);
 }
 
 .cd-button--light:hover svg,
 .cd-button--light:focus svg {
-  fill: var(--cd-primary-color--dark);
+  fill: currentColor;
```
rupl, line 290: love it! when possible this is IMO preferable to custom props and vars

components/cd-button/cd-button.html
```diff
@@ -21,49 +20,117 @@
...
+        <span class="cd-button__text">Button G</span>
+      </button>
+
+      <p></p>
+
 
       <button type="button" class="cd-button cd-button--export cd-button--icon">
         <span class="cd-button__text">Export</span>
-        <svg class="cd-icon cd-icon--arrow-down">
+        <svg class="cd-icon cd-icon--arrow-down" aria-hidden="true" focusable="false" width="16" height="16">
```
rupl, line 80: Not related to `cd-button` in particular, but I'm starting to wish for some shorthand to output an SVG from our sprite with all these props that get repeated.

```
<deco-svg icon="arrow-down">
```

That seems to be about all that is needed to expand into the full markup with class and `xlink:href` anchor thing.

components/cd-byline/cd-byline.css
```diff
@@ -0,0 +1,41 @@
+.cd-byline {
+  margin-top: 2rem;/* same as pseudo's padding-bottom */
```
rupl, line 2: I feel like a number with comment that it matches another number should be turned into a custom prop.

components/cd-byline/cd-byline.css
```diff
@@ -0,0 +1,41 @@
...
+.cd-byline__wrapper {
+  display: flex;
+}
+
+.cd-byline__image {
+  width: 4rem;
+  height: 4rem;
+  flex-grow: 0;
+  flex-shrink: 0;
+  flex-basis: 4rem;
```
rupl, line 23: I am generally seeing flex shorthand in most components. Since this isn't using calc and won't trip up IE11 using shorthand, should it be converted?

components/cd-card/cd-card.html
```diff
@@ -0,0 +1,156 @@
...
+<link rel=stylesheet href="../cd-utilities/cd-utilities.css" type="text/css" media=screen>
+<link rel=stylesheet href="../cd-typography/cd-typography.css" type="text/css" media=screen>
+<link rel=stylesheet href="../cd-button/cd-button.css" type="text/css" media=screen>
+<link rel=stylesheet href="../cd-read-more/cd-read-more.css" type="text/css" media=screen>
+<link rel=stylesheet href="cd-card.css" type="text/css" media=screen>
+
+<div class="cd-layout-container">
+  <div class="cd-container cd-grid cd-grid-3-col">
+
+    <article role="article" class="cd-card node--view-mode-card [ cd-flow ]">
```
rupl, line 13: I see five instances, but highlighting one: `.node--view-mode-card` seems to be a drupal-specific class and it isn't referenced in the CSS.

components/cd-card/cd-card.html
```diff
@@ -0,0 +1,156 @@
...
+
+      <div class="cd-card__container [ cd-flow ]">
+
+        <div class="cd-card__title">
+          <a href="#" rel="bookmark">Sed dapibus lectus sit amet adipiscing egestas mauris est viverra nibh</a>
+        </div>
+
+        <div class="cd-card__description node__content">
```
rupl, line 21: `node__content` also doesn't get referenced in the CSS. I feel like removing these is the safe move, to avoid having something from Drupal come along and pop some styles onto the card accidentally.

components/cd-card/cd-card.html
```diff
@@ -0,0 +1,156 @@
...
+
+    </article>
+
+    <article role="article" class="cd-card node--view-mode-card [ cd-flow ]">
+
+      <div class="cd-card__image">
+        <img src="https://via.placeholder.com/410x210" alt="card image example">
+      </div>
+
+      <div class="cd-card__container cd-align--center [ cd-flow ]">
```
rupl, line 78: the `.cd-align--*` classes are another utility that has an ambiguous origin. My assumption is that I will find a component  called `cd-XXX` if I see a class name starting with that.

In this case it might be as easy as adjusting the dashes:

```
.cd--align-center {}
```

The double-dash after CD tells me this is a top-level CD thing, and `align-center` is the name of that thing. It's technically a mis-use of the double-dash; I guess a true namespace for the utilities like `.cd-util__align-center` would be the "correct" solution but that's a lot to type and look at, and presents its own problems since the top-level `cd-util` is nowhere to be found.

components/cd-date/cd-date.html.twig
```diff
@@ -1,15 +1,11 @@
 {{ attach_library('common_design/cd-date') }}
 
-<div class="cd-date cd-date-wrapper">
-  <div class="cd-date-month-wrapper">
-    <span class="cd-date-month">Jul</span>
+<div class="cd-date">
```
rupl, line 3: Could the outer tag be `<time>` with `datetime` attribute to supply machine-readable output?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time

The suggestion assumes we have a machine-readable date when using this template to output data. If we don't then we actualy shouldn't use `<time>`:

>  If the element does **not** have a `datetime` attribute, **it must not have any element descendants**, and the datetime value is the element’s child text content.

components/cd-dropdown/cd-dropdown.js
```diff
@@ -0,0 +1,396 @@
...
+  document.addEventListener('click', handleClickAway);
+
+  document.addEventListener('click', handleEscape);
+  window.addEventListener('click', handleResize);
+  document.addEventListener('click', handleToggle);
+
+  // Initialize toggable dropdown.
+  initializeToggables();
+
+}
```
rupl, line 396: I feel like we're polluting global scope without much need to. Many functions have unique names, but a few don't, such as `toggle`.

The component JS can at least be contained to an IIFE with its own root object, and then define each method individually on the existing object. All the event listeners just need a prefix added to work the same way. The methods can call each other using that same prefix:

```js
(function iife() {
  // Our global
  var cdDropdown = {};

  // each method gets added to the global
  cdDropdown.initializeToggables = function initializeToggables() {
    console.log('initializeToggables');
  };
  cdDropdown.calledInternally = function calledInternally() {
    // just to demo who called
    console.log('calledInternally from ' + calledInternally.caller);
  };
  cdDropdown.setToggable = function setToggable() {
    console.log('setToggable');
    // call one of the peers
    cdDropdown.calledInternally();
  };

  // set an event listener
  document.addEventListener('click', cdDropdown.setToggable);

  // init
  cdDropdown.initializeToggables();
})();
```

Do other parts of the CD try to use these methods? If they do, we can ditch the `iife()` part and contain the whole thing to `window.cdDropdown`. It doesn't look like the code stores any state, but that can also be done directly inside `cdDropdown` with properties.

components/cd-event/cd-event.html
```diff
@@ -0,0 +1,173 @@
...
+
+    </article>
+
+
+    <article role="article" class="cd-card node--view-mode-card [ cd-flow ]">
+
+      <div class="cd-card__date--event">
+        <div class="cd-date">
+          <div class="cd-date__wrapper">
+            <span class="cd-date__day">18-20</span>
```
rupl, line 104: this one wraps like so on the demo:

```
18-
20
```

Is that intentional?

components/cd-dropdown/cd-dropdown.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr" class="no-js">
+  <head>
+    <meta charset="utf-8">
+    <title></title>
```
rupl, line 5: Just curious mainly, but most other HTML files are fragments that don't have `<html>`, `<head>`, `<body>` and so forth. Is this a special case?

components/cd-facets/cd-facets.css
```diff
@@ -0,0 +1,146 @@
+.block-facets {
```
rupl, line 1: I'm assuming faceted search markup is difficult and/or annoying to override, and that's why this component doesn't use `cd-facets` as a namespace for its classes?

components/cd-flow/cd-flow.css
```diff
@@ -0,0 +1,25 @@
...
+/* orientated towards simplicity, pragmatism and consistency */
+/* Composition Utility Block Exception */
+
+.cd-flow > * + * {
+  margin-top: var(--cd-flow-space, 1rem);
+}
+
+.cd-flow > .visually-hidden + * {
+  margin-top: 0;
+}
```
rupl, line 13: nice 💯

rupl: actually now that I think about it, if `.visually-hidden` is not the first element, it introduce weird spacing. It might be worth doing:

```css
.cd-flow > .visually-hidden:first-child + * {
  margin-top: 0;
}
```

components/cd-flow/cd-flow.css
```diff
@@ -0,0 +1,25 @@
...
+
+[dir='ltr'] .cd-bumper > * + * {
+  padding-left: var(--cd-bumper-space, 1rem);
+}
```
rupl, line 17: it's unclear to me what `cd-bumper` accomplishes. I see it used in `cd-author` and `cd-byline` but Flow is about vertical rhythm and Bumper seems to be about text directionality.

The demo page understandably doesn't have `cd-flow` as a dedicated component, so perhaps the README could include a few words about how bumper is intended to be used.

components/cd-alert/cd-alert.css
orakili: @rupl @left23 Unfortunately, the [css-vars-ponyfill](https://github.com/jhildenbiddle/css-vars-ponyfill) we use, only handles custom properties declared in `:root` so that would not work in IE11:

> Custom property declaration support is limited to :root and :host rulesets

components/cd-button/cd-button.html
orakili: If the idea is to avoid writing the full markup in templates, we could probably extend twig with a function that would output the markup like `{{ cd_svg_icon('arrow-down') }}`.

components/cd-date/cd-date.html.twig
orakili: I'm 💯 in favor of that. But, as often, Drupal is in the way and will strip out the `datetime` attribute... I've been bitten by this when working on ODSG and had to patch core: https://github.com/UN-OCHA/odsg8-site/blob/develop/patches/core--drupal--xss-prevent-protocol-stripping-on-datetime-attribute.patch. The list of accepted attributes is hardcoded so no way to deal with that in a `hook`.

components/cd-button/cd-button.css
left23: Ya, good point. I think the input selector should probably be removed. IIRC `input[type='submit']` came from my trying to remove the left-over SASS components from https://github.com/UN-OCHA/common_design/blob/master/sass/components/ which are there for Drupal and not always used.
I'll review this in more detail and see how removing it affects Drupal and whether I need to add something additional elsewhere to account for this.

components/cd-button/cd-button.html
left23: Adding as a backlog item

components/cd-alert/cd-alert.css
rupl: well 💩... it makes me sad, but the existing code works fine so we should leave it as-is.

Still, it doesn't sit right with me to use modern tools, but implement them in a way that caters to a browser that will never support the full potential. Maybe with [even MS ditching IE11](https://redmondmag.com/articles/2020/08/17/end-of-support-ie11-and-edgehtml.aspx) (article says MS Teams dropped support in Nov 2020, and Office365 ends in Aug 2021) we can start thinking about how to remove IE11 from our "feature parity" tier and downgrade it to something lower. I can't imagine OICT will allow it to stay around if it no longer receives security updates.

components/cd-card/cd-card.html
left23: These classes will be in other content-related components too. I'll remove the classes throughout anda also the `.node--view-mode` from the comment above.
I originally included them to make the mental link between the specific component and what we are used to in Drupal, but that isn't necessary.

components/cd-button/cd-button.html
rupl: @orakili yes, exactly the type of solution I was thinking of. my sample markup was pseudo-code.

We do have a Vue component to use as a model, which allows additional classes and a custom fill color: https://github.com/UN-OCHA/hid_v3/blob/dev/components/CommonIcon.vue

components/cd-date/cd-date.html.twig
rupl: I guess no change here then. Forcing core patches alongside CD is not cool IMO.

components/cd-card/cd-card.html
left23: Thanks, I take your points. Not sure how to proceed exactly but it's worth rethinking.

rupl: All my comments with `cd-utilities` might be "beginner" problems that go away with familiarity.

However that file contained multiple things that I had to go "figure out" as I reviewed the code. I suppose all the other components re so nicely scoped and contained, so it was jarring to have to go hunt down classes elsewhere. You spoiled me!

components/cd-dropdown/cd-dropdown.js
left23: @rupl There's a note in the README https://github.com/UN-OCHA/common_design/tree/cd-226-design-system/components/cd-dropdown#caveats where I try to explain why the JS file is here.
This one is not really used, but since I was testing things outside of Drupal, and might have to support these components outside of Drupal, I've left it in place and tried to keep it up-to-date with the work done on the one we use in Drupal https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/js/cd-dropdown.js
I'll do what I can to improve it

components/cd-event/cd-event.html
left23: Yes. This was initially to call attention to the design problem. I wanted to show the limitation of components in relation to their content, and demonstrate how proposed design do not accommodate. I'll change the fixed width to a min-width.

components/cd-dropdown/cd-dropdown.html
left23: Yep, this was so I could emulate Drupal's no-js / js class added to the html tag so the cd-dropdown.js in this component could stay as close as possible to the one used in Drupal (`common_design/js/cd-dropdown.js`)
https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/components/cd-dropdown/cd-dropdown.html#L9
https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/components/cd-dropdown/cd-dropdown.js#L384

components/cd-dropdown/cd-dropdown.js
rupl: whoops! sorry I missed the comment in the readme.

That's an interesting one. No fun to maintain two copies. Perhaps we should zero this file out and leave a comment saying "check the theme directory, or for non-drupal implementations remove the Drupal Behaviors wrapper and replace with a generic object"

components/cd-dropdown/cd-dropdown.html
rupl: gotcha 👍

components/cd-facets/cd-facets.css
left23: This is so Drupal-specific it almost shouldn't be part of the components as there will likely never be a plain HTML version of this. 
I'd consider this just underdeveloped so I can't comment specifically on whether overriding is easy or not. When I am updating the implementations, I will have to revisit this since many use Facets. I will look at it again from that perspective.

components/cd-flow/cd-flow.css
left23: Oops, ya, I didn't realise the README was essentially blank.
The `cd-bumper` is an attempt to control horizontal spacing. I bulked this in with `cd-flow` because I'm not sure how to solve the issue of consistent horizontal spacing, and it's a similar problem to vertical rhythm.
I'm unsure about whether this is the way to do it, and so I maintain it doesn't warrant its own `cd-bumper` component as I still might remove it altogether.
Also, `cd-flow` could be renamed to `cd-cubecss` or something if I need it to accommodate more rules.

components/cd-alert/cd-alert.css
left23: Yep, good point. I'm adding some feedback from Andy too
https://github.com/UN-OCHA/common_design/pull/141

```
/* If it's a warning or an error message, div should have role='alert' which is equivalent to aria-live='assertive'. */
.cd-alert div[role='alert'],
/* If it's a status message, div should have aria-live='polite' */
.cd-alert div[aria-live='polite'] {
```

components/cd-alert/cd-alert.css
left23: Thanks for elaborating. I've got a note about when/how we phase out IE11 support on the "code road map". To be continued.

components/cd-utilities/cd-utilities.css
left23: Yes, this is open for discussion. I've been using the utilities as a place for things I don't want to make into components because they are too small to warrant it, but I don't know where else they should go. I'm happy to rename this component or any selectors to improve clarity.

PR #118, opened 2020-11-04
[CD-233] Fix language switcher block html structure to ensure the creation of the toggle button

orakili:
Ticket: CD-233

TL;DR: v2.8.4 broke the language switcher. This attempts to fix the issue.

### Issue

In https://github.com/UN-OCHA/common_design/pull/108, to improve accessibility, a heading was added to the language switcher block. However the resulting html structure is not compatible with the `cd-dropdown.js` and is also not consistent with the other dropdown elements:

```html
<div class="cd-language-switcher">
  <h2>Language switcher</h2>
  <ul data-cd-toggable="en">
     ...
  </ul>
</div>
```

Expected (data-cd-toggable on the container):

```html
<div class="cd-language-switcher" data-cd-toggable="en">
  <h2>Language switcher</h2>
  <ul>
     ...
  </ul>
</div>
```

What happens is that the `<h2>` is detected as the potential "toggler" for the dropdown but because it's not a button, the `cd-dropdown.js` doesn't know how to deal with it and bails out (see [code](https://github.com/UN-OCHA/common_design/blob/eac3816e0af2e9f5ee4f3bfd96581be38754482c/js/cd-dropdown.js#L262-L282)).

### Solution

This PR moves the data-cd-toggable attributes to the parent of the `<ul>` (like it's done for the other "toggables" (main menu, search etc.)).

Screen Shot 2020-11-04 at 11 27 12
Screen Shot 2020-11-04 at 11 26 50

It also removes the `heading` code from the `links--language-block.html.twig` template to avoid having a double heading as the `<h2>Language switcher</h2>` is already added in the `block--language-block.html.twig` template.


left23 (approved):
@orakili Great, thanks.
I'm adding a selector to an existing CSS rule to visually hide the Block title if it's enabled.

PR #119, opened 2020-11-05
CD-234 Tailwind CSS framework

left23:
Tailwind CSS integrated with Common Design base theme

**Install TailwindCSS**
In base theme folder
`nvm use`
`npm install`

**To compile Tailwind**
`gulp tailwind`

There is example markup in page--demo.html.twig


left23:
Closing with no action as we decided this was overkill at the moment.

PR #120, opened 2020-11-06
CD-223 wrap strings in translation tags

left23:
Based on recent GHO work https://github.com/UN-OCHA/gho-site/pull/4/files

### DIFF
--- a/templates/cd/cd-footer/cd-copyright.html.twig
+++ b/templates/cd/cd-footer/cd-copyright.html.twig
@@ -1,8 +1,8 @@
 <div class="cd-footer__section cd-footer__section--copyright">
   <div class="cd-footer-copyright">
-    <span class="cd-footer-copyright__text">Except where otherwise noted, content on this site is licensed under a <a href="https://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution 4.0</a> International license.</span>
+    <span class="cd-footer-copyright__text">{{ 'Except where otherwise noted, content on this site is licensed under a <a href="https://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution 4.0</a> International license.'|t }}</span>
     <a class="cd-footer-copyright__link" href="https://creativecommons.org/licenses/by/4.0/">
-      <span class="visually-hidden">Creative Commons BY 4.0</span>
+      <span class="visually-hidden">{{ 'Creative Commons BY 4.0'|t }}</span>
       <svg class="cd-icon cd-icon--cc" width="32" height="32" aria-hidden="true" focusable="false">
         <use xlink:href="#cd-icon--creative-commons"></use>
       </svg>
--- a/templates/cd/cd-footer/cd-copyright.html.twig
+++ b/templates/cd/cd-footer/cd-copyright.html.twig
@@ -1,6 +1,6 @@
 <div class="cd-footer__section cd-footer__section--copyright">
   <div class="cd-footer-copyright">
-    <span class="cd-footer-copyright__text">{{ 'Except where otherwise noted, content on this site is licensed under a <a href="https://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution 4.0</a> International license.'|t }}</span>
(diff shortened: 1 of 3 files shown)


templates/cd/cd-footer/cd-copyright.html.twig
```diff
@@ -1,8 +1,8 @@
 <div class="cd-footer__section cd-footer__section--copyright">
   <div class="cd-footer-copyright">
-    <span class="cd-footer-copyright__text">Except where otherwise noted, content on this site is licensed under a <a href="https://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution 4.0</a> International license.</span>
+    <span class="cd-footer-copyright__text">{{ 'Except where otherwise noted, content on this site is licensed under a <a href="https://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution 4.0</a> International license.'|t }}</span>
```
rupl, line 3: Not a show-stopper, but when translating this one I noticed that not using Drupal's parameterization results in a weird UX on the backend. For example:

image

Now, I imagine that we might be the only people ever coming in contact with this issue, but I still wanted to highlight this specific case as weird when building GHO.

templates/cd/cd-footer/cd-mandate.html.twig
```diff
@@ -1,11 +1,11 @@
 <div class="cd-footer__section cd-footer__section--mandate">
   <div class="cd-mandate">
-    <span class="cd-mandate__heading">Service provided by</span>
+    <span class="cd-mandate__heading">{{ 'Service provided by'|t }}</span>
     <span class="cd-mandate__logo">
-      <span class="visually-hidden">UN-OCHA</span>
+      <span class="visually-hidden">{{ 'UN-OCHA'|t }}</span>
```
rupl, line 5: Bonus: you can close CD-235 by removing the hyphen.

templates/cd/cd-footer/cd-copyright.html.twig
left23: I assume this is happening with all languages? To do with the hyperlink I guess.
Any suggestions how to deal with it? I did some googling but nothing stands out, and I'm not sure what set up I need to test this.

rupl: Yes it's a general issue with translation, in no way specific to Arabic (but perhaps to our eyes more obvious how awkward it is to mix foreign characters with HTML).

When fixing this outside the theme system there is a way to parameterize strings such that the translation can be made dynamic, or to mix a translation with "unsafe" elements like HTML tags: https://getlevelten.com/blog/mark-carver/drupal-8-twig-templates-and-translations

```twig
<p class="submitted">{{ "Submitted by !author on @date"|t({ '!author': author, '@date': date }) }}</p>
```

If memory serves, the `!author` allows HTML, and `@date` will escape everything, thus is only suitable for text. It's tough to keep that syntax in your head if you're not doing translation all the time, so for that reason the same LevelTen article goes on to convert the above example into a `{% trans %}` block:

```twig
<p class="submitted">
{% trans %}
  Submitted by {{ author.username }} on {{ node.created }}
{% endtrans %}
</p>
```

We've used the `{% trans %}` block [in GHO](https://github.com/UN-OCHA/gho-site/blob/ef6e22169a6d435f17396d7d43e63411ff6e6340/html/themes/custom/common_design_subtheme/templates/fields/field--taxonomy-term--needs-and-requirements.html.twig#L48) when it fit. Try one of the approaches out and let's see which one fits better in your opinion.

left23: Ah ok, I didn't understand initially. The `{% trans %}` tag is familiar so I'm happy to go with that. I did some local testing and I think it works as I would expect, but please double check.

left23:
Screenshot of the UI translation:
Selection_006

@rupl @orakili If there are no objections, I will merge this.

rupl:
I'm personally ok with this for now. If we want to more accurately parameterize in the future, we can revisit.

PR #121, opened 2020-11-06
CD-224 xxl breakpoint

left23:
For use in Responsive image module and in SASS for larger viewport widths


left23:
Merging this. It shouldn't negatively impact.

PR #124, opened 2020-11-11
CD-211 OCHA Services in footer

left23:
Using js from @orakili from Reliefweb, we're loading the OCHA Services HTML in the footer and moving it to its location in the header with JS.
This means, if JS is disabled, or it content is being consumed other ways, it's in its a more semantic position in the source order. 

Best illustrated with JS disabled in browser (screenshot)

Before / After:
Selection_004


left23:
Some work needed. 

I added a visually-hidden h2 as well, as per comment https://github.com/UN-OCHA/common_design/pull/91 and I need to review and test.

PR #134, opened 2021-01-07
[CD-200] Header menus: fix flash of content

orakili:
Ticket: CD-200

The main goal of this PR is to fix the flash of dropdown menus that currently happens when the scripts are not cached.

To achieve this, the toggable menus are hidden until processed by the `cd-dropdown` script when javascript is enabled.

The menu templates have been simplified and consolidated adding also fallback elements (ex: links) that are replaced by the buttons when js is enabled.

The css rules for the button icons have also been a bit simplified.

There are 2 main changes to the `cd-dropdown` in the form of additional optional attributes:

- `data-cd-logo-only`: when set, the button label is visually hidden (ex: the help button)
- `data-cd-replace="ID"`: when set, the element with the given ID will be replaced by the toggler button. This is to allow fallback elements.

Lastly, there is a change in the `common_design.theme` to add tiny inline script in the `header` to be able to detect if javascript is enabled as early as possible to update the `no-js` class on the html because it's done quite late by the `Drupal.js` when not cached and that causes issues with css rules that depends on the `js` class being set.


left23:
@orakili This looks great, thanks for the work.
Can you give an example of a scenario where we would have a fallback element? I'm not sure what the intent is exactly.

left23 (approved):
Feel free to merge when you are happy with it.

orakili:
@left23 Regarding the fallback, I was thinking of progressive enhancement.

For example when the page is not consumed by a browser with javascript enabled, the **unprocessed** user menu may look like:

```html
<ul class="menu cd-user-menu">
  <li class="menu-item">
    <a href="/user/1234" id="cd-user-menu-item-0-1"><svg class="cd-icon--user">...</svg><span>Machin</span></a>
    <ul class="menu cd-user-menu__dropdown" data-cd-toggable="Machin" data-cd-replace="cd-user-menu-item-0-1">
        <li class="menu-item">
         <a href="/user/1234/subscriptions"><span>Subscriptions</span</a>
       </li>
       <li class="menu-item">
         <a href="/logout"><span>Log out</span</a>
       </li>
     </ul>
  </li>
</ul>
```

`<a href="/user/1234" id="cd-user-menu-item-0-1"><svg class="cd-icon--user">...</svg><span>Machin</span></a>` is the fallback that helps make sense of the menu and also access the user account page (otherwise, there would only be the `Subscription` and `Log out` links) when it's not transformed into the dropdown menu.

Another example could be to replace the **title** of a section. For example for the OCHA services:

```html
<section class="ocha-services">
  <h2 id="ocha-services-title"><svg class="cd-icon--ocha-logo">...</svg>OCHA Services</h2>
  <ul data-cd-toggable="OCHA Services" data-cd-replace="ocha-services-title" data-cd-logo="ocha--logo">
    <li>...</li>
    <li>...</li>
  </ul>
</section>
```

Would be transformed:

```html
<section class="ocha-services">
  <button data-cd-toggler><svg class="cd-icon--ocha-logo">...</svg><span>OCHA Services</span></h2>
  <ul data-cd-toggable="OCHA Services" data-cd-logo="ocha--logo">
    <li>...</li>
    <li>...</li>
  </ul>
</section>
```

It's a first step towards better progressive enhancement in my opinion though I think there still room for improvement, for example, copying the content of the fallback instead of limiting ourselves with the `data-cd-toggable`, `data-cd-logo` etc. attributes to create the `toggler` and maybe reviewing the use of the `aria-labelledby` for those toggable lists.

What do you think?

left23:
@orakili Thanks for elaborating and for the examples. Yes, it all sounds good to me.

For the first example, this is when there is only an icon for the dropdown button? Is there a site where this is currently the case when JS is disabled?

In the second example for the OCHA services, I understand this to mean we would do away with the `visually-hidden` class on the section title because when JS is disabled, we want the title to display, otherwise, it turns into a button and the title text becomes the button text. That's a nice improvement.

common_design.theme
```diff
@@ -165,6 +165,17 @@ function common_design_preprocess_html(&$vars) {
...
     $vars['attributes']['class'][] = 'path-error path-error--403';
   }
+
+  // Add a script to detect whether javascript is enabled or not and then
+  // replace the `no-js` class with the `js` one.
+  // Drupal.js does this but way too late on first load causing issues with
+  // css rules that rely on the `js` class being defined or not.
+  $js_detection = [
+    '#tag' => 'script',
+    '#value' => "document.documentElement.className = document.documentElement.className.replace('no-js', typeof document.addEventListener !== 'undefined' ? 'js' : 'no-js');",
```
rupl, line 175: 🤦 https://www.drupal.org/project/drupal/issues/2391025

rupl: In case it's not clear: I support your decision 💯  but really wish we could inline files. The only other option might be to do something like:

```
$js_detection = [
  '#tag' => 'script',
  '#value' => file_get_contents('path/to/file.js'),
];
```

For something so small I don't mind reading inside PHP quotes, but maybe for future if we want to expand this script, we can cheat using this method ☝️

orakili: Yep, I agree.

orakili:
I'm merging and I'll open another PR to port that to the v1 next week.

left23:
@orakili Thanks.
I'll do the PR against `master` if it's all the same to you. I can coordinate the updates and the release with the bug fix for #137 and there may be more fixes to go in.

PR #135, opened 2021-01-11
wip - components copied from GHO

lazysoundsystem:
Two poor choices of components to copy from GHO. For discussion in the CD meeting only, not for merging (in case there was any doubt).


left23:
@lazysoundsystem This should branch from `cd-226-design-system` - I think you should be able to change the base branch without any issue.

To add the component to the demo page, see https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/templates/page--demo.html.twig

1. Add your item to the Table of Contents https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/templates/page--demo.html.twig#L113
2. Add your component as an include: https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/templates/page--demo.html.twig#L176 
 - H3 name of component, followed by back-to-top component (returns to TOC)
 - component include
```
        <h3 class="cd-styleguide"><a id="cd-title-list">cd-title-list</a>
          {% include '@cd-components/cd-back-to-toc/cd-back-to-toc.html.twig' %}</h3>

        {% include '@cd-components/cd-title-list/cd-title-list.html.twig' %}
```
If there are variants, I am duplicating the markup in the twig template to display the variants. For the most part, the twig templates of existing components contain the markup from the html file only, with some of the wrapper divs removed as these are provided by Drupal's page template i.e. The twig templates are not specific to a Drupal build but for component display only. This is because I've been focusing on the HTML/CSS/JS side and not the implementation side. For now, I think this is fine to continue this way.
Selection_013

lazysoundsystem:
Thanks @left23 - I added to the right branch in the common-design-site demo I have set up locally, but got confused about which branch to make the PR against in this repo. I'll change that. As you also adapted this component, could you share your branch so I can compare them and see where I could have improved it?

lazysoundsystem:
Now based off the right branch - but it's still missing some of the improvements mentioned - add a title to the list, use colors and margins more consistent with other cd components, add a variant for the separator.

left23:
@lazysoundsystem 
The component I did from the GHO article title is already in the components folder https://github.com/UN-OCHA/common_design/tree/cd-226-design-system/components/cd-title-list *Note the branch, so you may or may not have seen it depending on what branch you are on.
Also can be seen here: https://feature.commondesign-unocha-org.ahconu.org/demo#cd-title-list

Regarding the branch, what I mean earlier was the base branch via Github - I just changed it now from master to `cd-226-design-system` and that reduces the commit history to just your component additions.
Although it was also relevant what branch you branched from :)

components/cd-article-list/cd-article-list.css
```diff
@@ -0,0 +1,68 @@
+.cd-article-list {
+  margin-top: 4.5rem;
+}
+
+.cd-article-list__title {
+  margin: 0 0 1.5rem 0;
+  color: #1f1f1f;
```
left23, line 7: This should use existing colour variables as a CSS custom property for consistency https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/sass/cd/_cd-variables.scss#L47
`color: var(--cd-grey--light)`

components/cd-article-list/cd-article-list.css
```diff
@@ -0,0 +1,68 @@
...
+.cd-article-list__list {
+  margin: 0;
+  padding: 0;
+  list-style: none;
+}
+
+.cd-article-list__list__item {
+  margin: 0;
+  padding: 0;
+  border: 1px solid #d8d8d8;
```
left23, line 22: As above, no hexidecimals

components/cd-article-list/cd-article-list.css
```diff
@@ -0,0 +1,68 @@
...
+.cd-article-list__list__item + .cd-article-list__list__item {
+  border-width: 0 0 1px 0;
+}
+
+.cd-article-list-link {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  padding: 1.25rem 0;
+  color: #1f1f1f;
```
left23, line 34: As above

components/cd-article-list/cd-article-list.css
```diff
@@ -0,0 +1,68 @@
...
+}
+
+.cd-article-list-link__arrow.cd-icon {
+  display: block;
+  flex: 0 0 auto;
+  width: 0.625rem;
+  height: 0.625rem;
+  margin: 0 0.25rem 0 1rem;
+  /* @todo should this be the same color as the links or use the UN blue? */
+  fill: #3a90d2; /* UN blue */
```
left23, line 62: As above

common_design.libraries.yml
```diff
@@ -38,6 +38,11 @@ cd-article:
     component:
       components/cd-article/cd-article.css: {}
 
+cd-article-list:
+  css:
+    component:
+      components/cd-article-list/cd-article-list.css: {}
+
```
left23, line 45: Because we are using CSS custom properties, each library needs the ponyfill set as a dependency so it loads in IE11.
See https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/common_design.libraries.yml#L70

common_design.libraries.yml
```diff
@@ -150,6 +155,11 @@ cd-hero--style-three:
   dependencies:
     - common_design/cd-hero
 
+cd-hr:
+  css:
+    component:
+      components/cd-hr/cd-hr.css: {}
+
```
left23, line 162: As above, if CSS custom properties are used.

components/cd-article-list/cd-article-list.html
```diff
@@ -0,0 +1,57 @@
+<link rel=stylesheet href="../cd-base.css" type="text/css" media=screen>
+<link rel=stylesheet href="../cd-flow/cd-flow.css" type="text/css" media=screen>
+<link rel=stylesheet href="cd-article-list.css" type="text/css" media=screen>
+
+<div class="cd-layout-container">
+  <div class="cd-container">
+    <ul class="cd-article-list cd-article-list--icon [ cd-flow ]">
+      <li>
+        <a href="#">
+          <svg class="cd-icon cd-icon--arrow-right" aria-hidden="true" focusable="false">
```
left23, line 10: These should probably have a width and height as per https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/common_design.libraries.yml#L70

components/cd-hr/README.md
```diff
@@ -0,0 +1,7 @@
+# Separator Component
+
+## Purpose and Usage
+It's a horizontal rule, copied from GHO - an exercise only.
```
left23, line 4: Probably a good exercise to fill this out - why would we need a horizontal rule? When and how should a content editor use this?

components/cd-hr/cd-hr.css
```diff
@@ -0,0 +1,7 @@
+.cd-hr {
+  border: none;
+  width: 100%;
+  height: 1px;
+  margin: 2rem auto;
+  background: #d8d8d8
```
left23, line 6: Replace hexidecimal

components/cd-hr/README.md
```diff
@@ -0,0 +1,7 @@
+# Separator Component
```
left23, line 1: Typically I have been using the same name for the component folder and for its name. This make it easier for everyone to know we are all talking about the same thing.
So maybe `cd-separator` as `cd-hr` implies an awareness of HTML elements?

left23 (commented):
@lazysoundsystem It's worth comparing the cd-article-list with the one I did (renamed cd-title-list) https://github.com/UN-OCHA/common_design/tree/cd-226-design-system/components/cd-title-list

Main differences are that I split the component. There is already a [cd-block-title](https://github.com/UN-OCHA/common_design/tree/cd-226-design-system/components/cd-block-title) component which means we don't need to include the title or its styles in this component.
I started the BEM naming on the list element itself with the idea that the surrounding context (I think we can start calling this the "composition") should not be included in this component.
I picture the implementation being a paragraph type (or similar) on a page with potentially multiple paragraph types of different content. The block titles will be consistent because they all use the same component, and the overall page composition (the vertical rhythm of the paragraph types or equivalent wrapper divs) can be dealt with as a whole, not as part of each individual component.
Let me know if you have thoughts on this.
CC @rupl @orakili @attiks @emmajane

components/cd-article-list/cd-article-list.css
```diff
@@ -0,0 +1,68 @@
...
+  .cd-article-list-link__title {
+    font-size: 1.125rem;
+    line-height: 2rem;
+  }
+}
+
+.cd-article-list-link__arrow.cd-icon {
+  display: block;
+  flex: 0 0 auto;
+  width: 0.625rem;
```
left23, line 58: Since we are using REM units (based on 16px), I think it makes sense to try and keep the increments easy to mentally translate to 16 so instead of 0.625rem (which btw is 10px) I would use 0.5 or .75 (8 or 12px). This also ensures we stick to a grid-like system and will result in better visual consistency, and then we don't have random 10px elements amongst mainly 4, 8, 12, 16px elements. This is most important when it comes to margins and padding (spacing) but I think it can also apply to dimensions.

left23:
@lazysoundsystem At your leisure, would you adjust this PR to be for the `cd-hr` component.
I'm labelling it as "proposed". 

It seems like a valid component to me, specifically (maybe) for page building situations when the Content Editor can add a variety of paragraph types.

@rupl @orakili What are your thoughts on this being a component? Can you summarise its purpose on GHO?
https://github.com/UN-OCHA/gho-site/tree/develop/html/themes/custom/common_design_subtheme/components/gho-separator

rupl:
> Can you summarise its purpose on GHO?

The GHO design called for lots of horizontal rules and after auditing the design plus discussing with designer, we ultimately decided there were no reliable conditions we could use to display them automatically.

We opted for a component simply to provide styling to an accompanying Paragraph type called "Separator" so that content editors could add the Separator wherever they wished. The Paragraph type was literally a drop-down asking "half or full width?" and no other input was possible.

> What are your thoughts on this being a component?

I think it's for a specific type of site like GHO, where it's supposed to look fancy and mimic print magazine (at least in the intent of the original GHO design). Do we want to promote that everywhere by offering it as a component? It might also be worth asking Javi during the next meeting if they find themselves using it often.

left23:
I think this can be closed.

lazysoundsystem:
Me too - it was useful for me as an exercise, but I don't think it needs to hang around any longer. Closing.

rupl:
One tiny follow-up: can the branch also be deleted?

left23:
Yep, I'll do it now.

PR #136, opened 2021-01-12
[CD-239] Port GHO figures components

orakili:
Ticket: CD-239

(Please ignore the wrong ticket number in the commit...)

This ports and merges the components used for the figures on the GHO site. Ex: https://gho.unocha.org/afghanistan

Screen Shot 2021-01-12 at 15 40 50

### Variants

There are 2 variants: `large` and `small` corresponding to the 2 lists of figures in the screenshot above.

### Caveats

The main caveat is that the `large` variant currently expects 3 figures. This can be altered by changing the `flex-basis: 33.33%;` rule.

This codepen shows some way to add some flexibility: https://codepen.io/heydon/pen/RvZQXR and variants could be added to handle different quantities with only a few rules.

### Screenshots

Screen Shot 2021-01-12 at 15 04 50
Screen Shot 2021-01-12 at 15 05 06
Screen Shot 2021-01-12 at 15 39 06


left23 (approved):
Thanks. I've made a few changes. This component is a nice addition.

PR #137, opened 2021-01-14
AR-208: Add cache context for user

attiks:


common_design.theme
```diff
@@ -20,6 +20,7 @@ function common_design_preprocess(&$variables, $hook) {
   // Add username.
   $user = \Drupal\user\Entity\User::load(\Drupal::currentUser()->id());
   $variables['username'] = $user->getUsername();
+  $variables['#cache']['contexts'][] = 'user';
```
orakili, line 23: I think this should be added to a more specific preprocess hook like `hook_preprocess_block()` and check the `$variables['plugin_id']` (or something like that) to only add this cache context to the `account` block. (Alternatively check the `$hook` variable).

Otherwise this cache context will be added to any rendered content which is probably not the expected result.

Similarly, I think the `username` should only be added in a `hook_preprocess_menu` as it's only used for the `menu__acount`.

orakili (changes requested):
See comment below, I think the cache context should be moved to another preprocess hook or be wrapped in some conditions to limit it to the `account` block.

common_design.theme
left23: @orakili @attiks I'm not sure what Jonathan's comments indicate. Does the cache and the username variable need to be added to separate preprocess functions?
I've added all the code to menu preprocess. On my local, I see no change i.e. the User menu still displays as expected.

left23 (approved):

common_design.theme
emmajane: @left23 My reading of Jonathan's comment is that the current code is too broad. By adding the additional line it will ONLY affect the username parts. To be fair though: I skipped the first and third paragraph and did some heavy guessing based on the second paragraph. (There were also some notes in [Flowdock](https://www.flowdock.com/app/unocha/digital-services/threads/so3EUGk24_WoNmcVvaw-ebbokG8).)

attiks: > I've added all the code to menu preprocess

Correct

orakili (approved):
Tested and it's working well, thanks.

PR #140, opened 2021-01-22
CD-250: port the GHO Appeals tag

rupl:
# CD-250

This PR creates a new `cd-tag` component which originated as [GHO Appeals tags](https://github.com/UN-OCHA/gho-site/tree/develop/html/themes/custom/common_design_subtheme/components/gho-appeals-tags).

### CSS

I made a few changes from GHO, because we had _extremely specific_ pixel-based requirements, so I tried to adapt the component to be a bit more flexible as is seen with the two other sizes.

### Template

The markup in GHO added classes by overriding the `taxonomy-term.html.twig` template with a vocab-specific one. I'm not sure if/how I should be providing a "base" override template. Here's the markup:

```twig
{%
  set classes = [
    'taxonomy-term',
    'vocabulary-' ~ term.bundle|clean_class,
    'cd-tag',
  ]
%}
{% set tagname = name|render|striptags|trim %}
<div{{ attributes.addClass(classes, ('cd-tag--' ~ tagname)|clean_class) }} title="{{ content.field_tag_extended|render|striptags|trim }}">
  {{- tagname -}}
</div>
```

You might spy a reference to `field_tag_extended` which is also an optional Extended name for a class, e.g. in GHO we have tags for things called Regional Refugee Appeals which get rendered RRP, but we wanted to provide an extended title for screen readers, or even curious mouse users who hover.

### Demo

My local shows something like this:

CD-250-example


left23:
@rupl Can you add everyone included in the CD dev calls as reviewer please

rupl:
done 👍  (except Javi?)

components/cd-tag/cd-tag.css
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * - margin is set in `rem` to ensure that multiple tag sizes visually line up
+ * - all other units are `em` so that setting `font-size` can scale them easily
+ */
+.cd-tag {
+  display: inline-block;
+  margin: 0.25rem;
+  padding: 0 .25em;
+  background-color: #5090cd; /* default is up for discussion */
+  color: #fff;
```
left23, line 19: Colours should use existing colour variables as a CSS custom property for consistency https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/sass/cd/_cd-variables.scss#L47

common_design.libraries.yml
```diff
@@ -198,6 +198,11 @@ cd-table:
     - common_design/css-vars-ponyfill
     - common_design/array-from-polyfill
 
+cd-tag:
+  css:
+    component:
+      components/cd-tag/cd-tag.css: {}
+
```
left23, line 205: Because we are using CSS custom properties, each library needs the ponyfill set as a dependency so it loads in IE11.
See https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/common_design.libraries.yml#L70

components/cd-tag/cd-tag.css
```diff
@@ -0,0 +1,75 @@
...
+.cd-tag--linked a:hover {
+  color: inherit;
+  text-decoration: none;
+}
+
+/**
+ * Variant: large
+ */
+.cd-tag--large {
+  font-size: 1em;
```
left23, line 67: REM units please

components/cd-tag/cd-tag.css
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/**
+ * Variant: huge
+ */
+.cd-tag--xl {
+  font-size: 1.5em;
```
left23, line 74: As above

components/cd-tag/cd-tag.css
```diff
@@ -0,0 +1,75 @@
...
+ * - margin is set in `rem` to ensure that multiple tag sizes visually line up
+ * - all other units are `em` so that setting `font-size` can scale them easily
+ */
+.cd-tag {
+  display: inline-block;
+  margin: 0.25rem;
+  padding: 0 .25em;
+  background-color: #5090cd; /* default is up for discussion */
+  color: #fff;
+  font-size: .625em;
```
left23, line 20: REMs please. 
I would bump this up to 0.75rem or 1rem until a designer says otherwise.

components/cd-tag/cd-tag.css
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * Individual tag
+ *
+ * - margin is set in `rem` to ensure that multiple tag sizes visually line up
+ * - all other units are `em` so that setting `font-size` can scale them easily
+ */
+.cd-tag {
+  display: inline-block;
+  margin: 0.25rem;
+  padding: 0 .25em;
```
left23, line 17: REMs please

left23 (commented):
Thanks @rupl
I would be curious to make the description in the README and the CSS file a bit less Drupal-specific in the explanation, and see how it might be understood outside of the Drupal context.
I'm also wondering how we'd avail of the OCHA colour scheme to provide a good starting point and/or more options without having to customise.
We have a lot of colours we could avail of https://brand.unocha.org/d/xEPytAUjC3sH/visual-identity#/basics/colours/secondary-colours but currently no real colour system as such (most of the colours are not included as CSS custom props in the theme at the moment)
I've been contemplating how/if to integrate the colours more but I haven't arrived at anything concrete. This is so far the best candidate to try something.
The colours have english names, maybe we could use those. Or maybe we just expand the variants you started:
```
.cd-tag--purple,
.cd-tag--variant-3 {
        background-color: var(--cd-purple)
      }
```
To be continued.

components/cd-tag/cd-tag.css
rupl: Per the comment block, converting this particular unit to `rem` could result in a disproportional-looking component when the different sizes are used. I set up the demo page to show this exact effect (or rather, how nice it can look with many different-sized tags)

But if `em` truly has to go, then I have another idea to further componentize this.

components/cd-tag/cd-tag.css
rupl: This also destroys a "feature" of the component, which allows you to tag content of various font-sizes and have it automatically retain proportion. On GHO we had tags on page titles, but on HID it's on body-copy-sized text, etc.

left23: @rupl I meant to get back to this. I'm not suggesting we destroy the feature of the component at all. I didn't understand the details. This seems worth rewinding.

left23:
@rupl I see the colours matching https://brand.unocha.org/d/xEPytAUjC3sH/visual-identity#/basics/colours and them maybe some different shades depending on what works with the white text colour? If so let's include this in the readme - a link to the brand colour page.

I like the system and I think this works well.
The extra classes `taxonomy-term vocabulary-tags` aren't necessary but maybe help link this to Drupal? I'm not so familiar with Drupal taxonomy.

Can you share a link to where on GHO these appear. I'm curious to see how they fit in with elements around them.
For example, do you picture multiple tags in one line and then a heading beneath (or reverse?), or a heading next to a tag?
```
<h2>The heading with the tag</h2>
<div class="cd-tags">
  <div class="taxonomy-term vocabulary-tags cd-tag cd-tag--rounded cd-tag--color-4" title="Rounded Tag">Rounded</div>
  <div class="taxonomy-term vocabulary-tags cd-tag cd-tag--rounded cd-tag--color-5" title="Rounded Tag 1">Rounded 1</div>
</div>
```

Are you happy with this component as is for the use case we have for the HID labels?

rupl:
Here's a page with several tags: https://gho.unocha.org/inter-agency-appeals/latin-america-and-caribbean — look for major headings with country names: Colombia, Haiti, etc.

I will put the `em` back because it IMO provides a more out-of-the-box usefulness.

rupl:
The class names having to do with taxonomy have also been scrubbed. I agree with you: for a design system let's stick with our component class names, and do a reset on individual implementations (or elsewhere inside the base theme if down the road we find it coming up often).

rupl:
Oops one more question to answer: yes I like this for HID 👍

PR #141, opened 2021-01-25
CD-226 cd-alert

left23:
Review the cd-alert component

### DIFF
--- a/components/cd-alert/cd-alert.css
+++ b/components/cd-alert/cd-alert.css
@@ -6,7 +6,7 @@
 }
 
 .cd-alert,
-.cd-alert div[role='alert']{
+.cd-alert div[role='alert'] {
   display: flex;
   align-items: flex-start;
 }
--- a/components/cd-alert/cd-alert.css
+++ b/components/cd-alert/cd-alert.css
@@ -3,6 +3,9 @@
   padding: 1rem;
   margin: 1rem 0 2rem;
   border-radius: 0.25rem;
+  border: 1px solid var(--cd-blue--bright);
+  box-shadow: -8px 0 0 var(--cd-blue--bright);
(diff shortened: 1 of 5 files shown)


components/cd-alert/cd-alert.css
```diff
@@ -3,10 +3,13 @@
...
   margin: 1rem 0 2rem;
   border-radius: 0.25rem;
+  border: 1px solid var(--cd-blue--bright);
+  box-shadow: -8px 0 0 var(--cd-blue--bright);
+  color: var(--cd-grey--dark);
 }
 
 .cd-alert,
-.cd-alert div[role='alert']{
+.cd-alert div[role='alert'] {
```
lazysoundsystem, line 12: Is this line necessary? Also, https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role suggests that `role='alert'` should be used for warning and error messages in the templates, but not 'status'.

Maybe: `aria-live="polite"` instead?

lazysoundsystem (approved):
All looks good. The icon names prompted me to try to think of a different name for this component, but 'alert' seems right.

components/cd-alert/cd-alert.css
left23: @lazysoundsystem Thanks. I've added a comment to the CSS to explain that selector, and I've adjust the markup to be inline with the actually output of the message template override https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/templates/status-messages.html.twig (comes from Classy)
I'm not sure about the aria-live attribute. I wonder if this is already taken into account in Drupal (I'm not sure how to output an error message in the frontend) so more investigation needed here.

PR #143, opened 2021-01-27
CD-226: explicitly set line-height to normalize's default value

rupl:
# CD-226 line-height `.cd-button`

While working on [HID-2169](https://humanitarian.atlassian.net/browse/HID-2169), I was using CD Buttons on both `<button>` and `<a>` elements which are horizontally aligned on a single row. The reason is because two buttons are submit/reset for a form, and the third was a hyperlink labeled "Cancel"

When I loaded the form, the Cancel button had more vertical space, leading the button to be taller. It ended up being a different `line-height` for links versus buttons. The PR further normalizes the value from `normalize.css` which sets buttons to `line-height: 1.15` and ensures that button-looking links also end up the same visual size.

### Before

Screen Shot 2021-01-27 at 17 01 55

### After

Screen Shot 2021-01-27 at 17 02 16

### DIFF
--- a/components/cd-button/cd-button.css
+++ b/components/cd-button/cd-button.css
@@ -49,6 +49,7 @@
   border: 2px solid transparent;
   padding: 0.5rem 1rem;
   font-size: 1rem;
+  line-height: 1.15;
   text-align: center;
   transition: background-color 0.6s ease-out,
   color 0.6s ease-out;


left23 (approved):
@rupl Well spotted, thanks for the PR.
Approving and merging.

PR #144, opened 2021-01-28
CD-244 webfonts

left23:
Add Roboto and Noto as per brand guidelines.
https://brand.unocha.org/d/xEPytAUjC3sH/visual-identity#/basics/fonts/advanced-users
https://brand.unocha.org/d/xEPytAUjC3sH/visual-identity#/basics/fonts/multi-language-typefaces-1

Roboto is there but commented out. It's optional and I guess this means those rules would need to be copied into the sub theme.

Noto is used for RU, ZH.
Noto Kufi is used for AR (see comments below)

See [CD-244](https://humanitarian.atlassian.net/browse/CD-244) for more.


sass/cd/_cd-variables.scss
```diff
@@ -120,9 +120,23 @@ $cd-primary--dark: $cd-blue--dark;
...
 $cd-font--system: system-ui;
 
-// Web fonts.
-//@import url('https://fonts.googleapis.com/css?family=Mali:300,400,500,600,700&display=swap"');
-//$cd-font--webfont: 'Mali', cursive;
+// sass-lint:disable no-url-domains no-url-protocols
+
+// Multilingual
+// https://brand.unocha.org/d/xEPytAUjC3sH/visual-identity#/basics/fonts/multi-language-typefaces-1
+@import url('//fonts.googleapis.com/css2?family=Noto+Sans:wght@400;700&display=swap');
```
lazysoundsystem, line 127: An argument to self-host the fonts: https://wicki.io/posts/2020-11-goodbye-google-fonts/

lazysoundsystem (approved):
Looks right, but there is a discussion to be had about self-hosting the fonts.

sass/cd/_cd-variables.scss
rupl: end of an era, eh?

Thanks for the link because otherwise I'd argue that people bouncing around between different OCHA sites would benefit from one cached resource.

We don't have to creep on this PR. It can be a separate ticket IMO.

left23: This is why the PR is here. We can get into it.

If I set this up for self-hosting, I would need to include the fonts in the repo, which I wanted to avoid doing because maintenance (maybe there is no maintenance? How often do fonts get updated?)

My first purpose here was to provide guidance for the fonts, and align with the brand guidelines.

If/when Roboto is used, that implementation will need to be in the subtheme so the base theme in this case serves only to provide some copy and paste rules, and take the thinking out of what font we should use. And because Roboto is optional, I also didn't want to include the font files.

As long as this is not blocking implementations from self-hosting fonts, I think it's a good first step. If we find we have a handful of implementations all opting for self-hosting, then maybe that's the evidence we should be more opinionated in the base theme.

What do you think?

rupl: > How often do fonts get updated?

Roboto is pretty stable nowadays. I think in three years I did it once for DSR.

> My first purpose here was to provide guidance for the fonts, and align with the brand guidelines.

One thing that Google Fonts supports nowadays (and I see you implemented already) is the `font-display: swap` — a default that is far more important than the origin of the font in most cases. If we set up self-hosting, we would be keen to provide the `@font-face` definitions for folks as well, to encourage that they get loaded uniformly across properties.

> As long as this is not blocking implementations from self-hosting fonts, I think it's a good first step.

Agreed, it is better to provide a recommendation than to let people go nuts.

## UNOCHA-hosted?

Third option that potentially allows us to have our 🍰  and eat it too?

If we set up a simple static repo, e.g. a GitHub Pages with a CNAME like `fonts.unocha.org` I wonder if that allows at least the sites under the unocha.org domain to gain the benefits of cross-site caching, which also reduces our need to re-implement self-hosted fonts over and over again.

If we build self-hosting into the theme it might be overkill to offer such a service, but the idea popped into my head and I wanted to jot it down.

left23: I do enjoy cake.
I'm adding a note on the JIRA issue to track this.

sass/cd/_cd-variables.scss
```diff
@@ -120,9 +120,23 @@ $cd-primary--dark: $cd-blue--dark;
...
+$cd-font--noto: 'Noto Sans', sans-serif;
```
rupl, line 128: Starting a new thread about Noto Sans in particular since the line containing the URL was the basis for the self-hosted versus CDN discussion.

This particular Noto font is "the Sans Latin, Greek and Cyrillic family." — Arabic and Chinese are not included in this font.

For DSR, we were advised by DDMU to use [Noto Kufi for Arabic](https://www.google.com/get/noto/#kufi-arab) for titles, and Dubai for body copy, both of which are **not** available on CDNs. You can see my [declarations on DSR here](https://github.com/UN-OCHA/reports-site/blob/dev/layouts/default.vue#L249-L271).

For GHO we used Kufi for all text, and did not use Dubai. Coincidentally, their Arabic page is live as of today so you can inspect here: https://gho.unocha.org/ar

I have personally never installed Chinese-compatible fonts so I'd probably start by asking DDMU what their go-to font is before we pick one of the 10 or so options available on the Noto site.

rupl: AFAIK performance-wise the browsers have most of this stuff figured out, as long as we aggressively scope our selectors defining font usage with `[lang='XX']` prefixes. That is, the fonts will not be downloaded until the CSSOM matches an actual DOM element worth triggering said download.

sass/cd/_cd-variables.scss
rupl: One more thought: our own non-unocha.org sites could still use this no different than google fonts. So implementation-wise it would be consistent across all our sites, with a potential performance benefit for unocha.org subdomains.

sass/cd/_cd-variables.scss
left23: @rupl Great, thanks for elaborating. I will follow up.

left23:
I've used the implementation of Noto Kufi from GHO for AR and adjusted the selectors to suit.
I've left a TODO for ZH as it will need work once we have a real-life situation that requires this.

https://github.com/UN-OCHA/gho-site/blob/49f9a6fc7a35dd71b70e509436585420424d189b/html/themes/custom/common_design_subtheme/templates/html.html.twig#L49

sass/cd/_cd-variables.scss
attiks: For foreign char sets we once used this

```css
/*Chinese*/

/*Traditional*/
@font-face {
  font-family: 'NotoChineseTraditionalRegular';
  font-display: swap;
  src: url('/fonts/NotoSansCJKtc-Regular-Alphabetic.woff2') format('woff2'), url('/fonts/NotoSansCJKtc-Regular-Alphabetic.woff') format('woff'), url('/fonts/NotoSansCJKtc-Regular-Alphabetic.ttf') format('truetype');
}

@font-face {
  font-family: 'NotoChineseTraditionalBold';
  font-display: swap;
  src: url('/fonts/NotoSansCJKtc-Bold-Alphabetic.woff2') format('woff2'), url('/fonts/NotoSansCJKtc-Bold-Alphabetic.woff') format('woff'), url('/fonts/NotoSansCJKtc-Bold-Alphabetic.ttf') format('truetype');
}

/*Simplified*/
@font-face {
  font-family: 'NotoChineseSimplifiedRegular';
  font-display: swap;
  src: url('/fonts/NotoSansCJKsc-Regular-Alphabetic.woff2') format('woff2'), url('/fonts/NotoSansCJKsc-Regular-Alphabetic.woff') format('woff'), url('/fonts/NotoSansCJKsc-Regular-Alphabetic.ttf') format('truetype');
}

@font-face {
  font-family: 'NotoChineseSimplifiedBold';
  font-display: swap;
  src: url('/fonts/NotoSansCJKsc-Bold-Alphabetic.woff2') format('woff2'), url('/fonts/NotoSansCJKsc-Bold-Alphabetic.woff') format('woff'), url('/fonts/NotoSansCJKsc-Bold-Alphabetic.ttf') format('truetype');
}

/*Japanese*/
@font-face {
  font-family: 'NotoJapanRegular';
  font-display: swap;
  src: url('/fonts/NotoSansCJKjp-Regular-Alphabetic.woff2') format('woff2'), url('/fonts/NotoSansCJKjp-Regular-Alphabetic.woff') format('woff'), url('/fonts/NotoSansCJKjp-Regular-Alphabetic.ttf') format('truetype');
}

@font-face {
  font-family: 'NotoJapanBold';
  font-display: swap;
  src: url('/fonts/NotoSansCJKjp-Bold-Alphabetic.woff2') format('woff2'), url('/fonts/NotoSansCJKjp-Bold-Alphabetic.woff') format('woff'), url('/fonts/NotoSansCJKjp-Bold-Alphabetic.ttf') format('truetype');
}

/*Greek*/
@font-face {
  font-family: 'NotoGreekRegular';
  font-display: swap;
  src: url('/fonts/NotoGreekRegular.woff2') format('woff2'), url('/fonts/NotoGreekRegular.woff') format('woff'), url('/fonts/NotoGreekRegular.ttf') format('truetype');
}

@font-face {
  font-family: 'NotoGreekBold';
  font-display: swap;
  src: url('/fonts/NotoGreekBold.woff2') format('woff2'), url('/fonts/NotoGreekBold.woff') format('woff'), url('/fonts/NotoGreekBold.ttf') format('truetype');
}

/*Russia*/
@font-face {
  font-family: 'NotoRussianRegular';
  font-display: swap;
  src: url('/fonts/NotoRussianRegular.woff2') format('woff2'), url('/fonts/NotoRussianRegular.woff') format('woff'), url('/fonts/NotoRussianRegular.ttf') format('truetype');
}

@font-face {
  font-family: 'NotoRussianBold';
  font-display: swap;
  src: url('/fonts/NotoRussianBold.woff2') format('woff2'), url('/fonts/NotoRussianBold.woff') format('woff'), url('/fonts/NotoRussianBold.ttf') format('truetype');
}
```

sass/cd/_cd-variables.scss
attiks: > GitHub Pages with a CNAME

This might work, but first request will be slow as well and we depend on another third party for hosting it.

rupl: Yup, well noted. The custom origin was just an idea for an experiment, and I don't think there's motion to implement atm.

sass/cd/_cd-variables.scss
rupl: good to know!

PR #146, opened 2021-01-28
CD-226 cd-read-more

left23:
Review the `cd-read-more` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-read-more


lazysoundsystem (approved):
Looks good.

PR #148, opened 2021-01-28
CD-226 cd-byline

left23:
Review the `cd-byline` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-byline


lazysoundsystem (approved):
Looks good

PR #150, opened 2021-02-04
[CD-257] Make the see all link label translatable in the OCHA services

orakili:
Ticket: CD-257

Simply makes the "See all" link label translatable.


left23 (approved):
Thanks @orakili
I will add this to `cd-226-design-system` branch too

PR #151, opened 2021-02-04
[CD-255] Fix menu attributes and add compatibility for GHO

orakili:
Ticket: CD-255

This does 2 thinks: 

1. Reset the "attributes" variable before passing it to the nested menus.

This prevents attributes added to the top menu elements to be added to the nested menus which is the equivalent of what `stable` and `classy` themes for example do via a `if/else`:

```twig
    {% if menu_level == 0 %}
      <ul{{ attributes }}>
    {% else %}
      <ul>
    {% endif %}
```

2. Add `span` around the `title` when displaying links in the navigation menu. This is for compatibility for GHO and ease styling (ex: centering the title vertically).

### DIFF
--- a/templates/navigation/menu--account.html.twig
+++ b/templates/navigation/menu--account.html.twig
@@ -78,12 +78,13 @@
         {% if item.is_expanded and item.below %}
 
           {%
-            do attributes
-              .setAttribute('data-cd-toggable', title)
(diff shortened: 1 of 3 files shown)


left23 (approved):
@orakili great, thanks.
I'll add this to `cd-226-design-system` branch too

PR #154, opened 2021-02-04
CD-226 cd-hero

left23:
Review the cd-hero component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-hero

### DIFF
--- a/components/README.md
+++ b/components/README.md
@@ -23,7 +23,7 @@ Add a check if you've reviewed the component and it passes the Criteria.
 | cd-filter           |                     |                     |
 | cd-flow             |                     |                     |
 | cd-grid             |                     |                     |
-| cd-hero             |                     |                     |
+| cd-hero             |       :x:           |                     |
 | cd-image-grid       |                     |                     |
 | cd-link-list        |                     |                     |
 | cd-pagination       |                     |                     |


lazysoundsystem (approved):
Some general comments:

As with cd-banner, do we need cd-bleed for the non-twig version? And should the heights be the same? 370px for banners, 256px or 450px for heros.

Is it worth trying for more descriptive names for the styles? (e.g 'gradient', 'big-block', 'inset')

And I don't think we need captions for heros, but we probably will need photo credits. Should these go in the caption area, or do we need a 'cd-photo-credit' component?

left23:
> As with cd-banner, do we need cd-bleed for the non-twig version? 
Yep, added it for clarity.

> And should the heights be the same? 370px for banners, 256px or 450px for heros.
> 
> Is it worth trying for more descriptive names for the styles? (e.g 'gradient', 'big-block', 'inset')
We discussed this initially during our CD call in (November 10th, feel free to browse notes) and decided we'd stick to the generic naming. If/when we provide a catalogue it will be clearly marked so there should be no confusion even if the variant names aren't using descriptors that match the visual style. The main reason we opted not to bother naming them was it was hard and we could talk forever about what different names mean to people. I'll hold to that decision for now. 
BTW I like your naming suggestions.

> And I don't think we need captions for heros, but we probably will need photo credits. Should these go in the caption area, or do we need a 'cd-photo-credit' component?

The caption is optional (and came from IASC wanting a caption for their Hero image). I am treating the credit and caption as the same thing. Typically, I have seen both the caption and credit combined:
https://interagencystandingcommittee.org/results-group-2-accountability-and-inclusion
https://www.unocha.org/story/daily-noon-briefing-highlights-burkina-fasosenegal-ethiopia-niger

Unless there are other examples that contradict this, I think for simplicity we stick to the one component. I'm wondering now if the name is confusion but let's get evidence on this before we bother changing this. For now, I've updated the README for cd-banner and cd-hero "Use with cd-caption component for image caption and/or image credits."

PR #155, opened 2021-02-04
CD-226 cd-banner

left23:
Review of the cd-banner component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-banner for examples

### DIFF
--- a/components/README.md
+++ b/components/README.md
@@ -8,7 +8,7 @@ Add a check if you've reviewed the component and it passes the Criteria.
 | cd-alert            | :heavy_check_mark:  |                     |
 | cd-article          |                     |                     |
 | cd-author           |                     |                     |
-| cd-banner           |        :x:          |                     |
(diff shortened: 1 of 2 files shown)


lazysoundsystem (approved):
Should we be using `srcset` to make sure we're not sending outsize images to mobile phones?

Other minor comments on https://github.com/UN-OCHA/common_design/commit/df2987ba7ee95291d910e58276ef3fa8c5933e13 and https://github.com/UN-OCHA/common_design/commit/f0f8296ffaed6090ab13ad4f9d28a7f7f969f626

left23:
> Should we be using `srcset` to make sure we're not sending outsize images to mobile phones?

Yes, but I was leaving this as an implementation detail. Definitely in Drupal we'll be using the Responsive image module which will handle all this.
It would be good to demonstrate this in the HTML too though as it might be a good starting point for the correct dimensions needed for the Responsive image module config. Time-permitting I will revisit.

PR #156, opened 2021-02-04
CD-226 cd-block-title

left23:
Review of the `cd-block-title` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-block-title for an example


lazysoundsystem (approved):
:heavy_check_mark:

PR #157, opened 2021-02-04
CD-226 cd-link-list

left23:
Review of the `cd-link-list` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-link-list for an example

### DIFF
--- a/components/README.md
+++ b/components/README.md
@@ -25,7 +25,7 @@ Add a check if you've reviewed the component and it passes the Criteria.
 | cd-grid             |                     |                     |
 | cd-hero             |                     |                     |
 | cd-image-grid       |                     |                     |
-| cd-link-list        |         :x:         |                     |
+| cd-link-list        | :heavy_check_mark:  |                     |
 | cd-pagination       |                     |                     |
 | cd-read-more        |                     |                     |
 | cd-search           |                     |                     |
--- a/components/README.md
+++ b/components/README.md
@@ -25,7 +25,7 @@ Add a check if you've reviewed the component and it passes the Criteria.
 | cd-grid             |                     |                     |
 | cd-hero             |                     |                     |
 | cd-image-grid       |                     |                     |
-| cd-link-list        |                     |                     |
+| cd-link-list        |         :x:         |                     |
 | cd-pagination       |                     |                     |
 | cd-read-more        |                     |                     |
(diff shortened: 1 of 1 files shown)


lazysoundsystem (approved):
On a wide screen, the chevrons aren't clearly connected with the text (until you notice them). I recently tried this https://christianheilmann.com/2020/11/26/back-to-basics-creating-a-clickable-card-interface-in-plain-html-css-and-javascript/ and it's very simple and effective and could be useful here.

I noticed that the html template has more links than the twig one, and includes cd-grid.

And for consistency, we could have an 'rtl' section to flip the chevron.

left23:
> On a wide screen, the chevrons aren't clearly connected with the text (until you notice them). I recently tried this https://christianheilmann.com/2020/11/26/back-to-basics-creating-a-clickable-card-interface-in-plain-html-css-and-javascript/ and it's very simple and effective and could be useful here.

I know what you mean. This is based on the design. I think the original intent was for this to be a sidebar component so the content and the icon were meant to be close together (screenshot from Figma mockups)
Selection_010
However, there is nothing preventing this being used in a full-width situation so at the very least we should make a note of this in the README.

I personally don't like the pattern where an entire teaser is clickable because I feel it's misleading to the user and promotes accidental clicking and ending up on different pages without being completely aware as to why - especially on mobile where the white space is often used for touching to scroll. I don't think it warrants it in this case, as we don't really have good evidence this pattern actually improves UX. We're going for simplicity where possible and for me that starts with HTML elements acting like we'd expect. If/when we have evidence where this is not adequate we should revisit.

> I noticed that the html template has more links than the twig one, and includes cd-grid.
Yep, I will make sure these match by adding that second instance to the Twig template too. The ideas was to see how easily I could apply a grid layout to this component. It seems to work alright so it's good to demonstrate. It was came from trying to figure out how to deal with this component when it displays the full width of the main div.

Selection_011



> And for consistency, we could have an 'rtl' section to flip the chevron.
Yes, there's a lot of work needed for proper RTL but this is a good start and we've got similar rules in other components. I'll add it.

left23:
I will spend a little time working with this component and the `cd-title-list` to see whether I can easily go from just a title link and chevron icon, to image, title, description text and chevron icon as a variant option.

PR #158, opened 2021-02-04
CD-226 cd-title-list

left23:
Review of the `cd-title-list` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-title-list for an example


lazysoundsystem (approved):
Wondering if this could be part of cd-link-list.

left23:
> Wondering if this could be part of cd-link-list.

I've wondered too but I like the simplicity of this one. It doesn't allow for any other info, whereas `cd-link-list` is to cater for a 
variety of content displays. I will comment more about that one on its PR.

PR #159, opened 2021-02-08
CD-226 cd-link-list vs cd-title-list

left23:
As described on Trello https://trello.com/c/GEMMqZ60/1-cd-title-list-vs-cd-link-list
first pass to combine these two components.
Not really sold on the idea so leaving it here and setting this as draft.


left23:
I will close this unless there's objections. It illustrates two ways of doing the same thing, and the component never became relevant enough for this to move forward.

PR #160, opened 2021-02-08
CD-226 cd-card

left23:
Review of the `cd-card` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-card for an example

### DIFF
--- a/components/README.md
+++ b/components/README.md
@@ -15,7 +15,7 @@ Add a check if you've reviewed the component and it passes the Criteria.
 | cd-button           | :heavy_check_mark:  |                     |
 | cd-byline           | :heavy_check_mark:  |                     |
 | cd-caption          | :heavy_check_mark:  |                     |
-| cd-card             |                     |                     |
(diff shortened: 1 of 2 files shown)


left23:
I noticed in `cd-event` component there should probably be a variant listed on the `cd-card` component as described in README https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/components/cd-event/README.md

lazysoundsystem (approved):
Added 1px to the media query for consistency - if it breaks something, do change it back.

Thinking again about making the whole card clickable, which maybe has more of a case here than in https://github.com/UN-OCHA/common_design/pull/157 - I also found this implementation, which has some other suggestions too: https://inclusive-components.design/cards/

But as it is it works well and is intuitive.

left23:
@lazysoundsystem Should there be a commit from you on this branch?

lazysoundsystem:
Yes, not sure what happened there. Also, edited my last comment - the media query change was for another component.

PR #161, opened 2021-02-08
CD-226 cd-styled-list

left23:
Review of the `cd-styled-list` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-styled-list for an example


lazysoundsystem (approved):
`@todo` in README is half done - there's a grid in the .html file but not the .twig one (and should the .html one be with `ul` and `li` rather than `div`s?)

Suggested changes are whitespace only but for the 1px change in the media query for consistency.

(And again, there's a lot of common ground between this and the other cd-*-lists.)

PR #162, opened 2021-02-08
CD-226 cd-teaser

left23:
Review of the `cd-teaser` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-teaser for an example


lazysoundsystem (approved):
I added the grid from the .html template to .twig and got this screenshot:
image
Looks like we need to add some margin for grid elements and come up with a solution for date ranges.

I don't know how often teasers would be used for events, but we could use icons for time and location.

left23:
I would disregard anything in the html that didn't make it into the corresponding twig file. It usually meant I'd given up or decided against something but forgot to update the html file. Please do ask questions, but typically the twig is the authority because that's what I'm checking via the /demo page template override in Drupal.

In short, while it might be nice to have the teasers work in a grid, it's not currently catered for because a.) maybe in that case they should actually be cards, and b.) I have no evidence this will be useful yet. We can do this on-demand.

I agree the icons would make a nice addition. 

Thanks. I'll add both points to the readme.

PR #163, opened 2021-02-08
CD-226 cd-date

left23:
Review of the `cd-date` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-date for an example


lazysoundsystem (approved):
Less concise, but should it be `cd-date-wrapper` be `cd-date-date-wrapper` to be consistent with `cd-date-day-wrapper`?

left23:
> Less concise, but should it be `cd-date-wrapper` be `cd-date-date-wrapper` to be consistent with `cd-date-day-wrapper`?

I've refactored this to have proper BEM classes. Thanks for pointing it out.

PR #164, opened 2021-02-08
CD-226 cd-event

left23:
Review of the `cd-event` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-event for an example


lazysoundsystem (approved):
Noticed the links are different colors, depending on whether they are direct descendants of `__title` elements or not:
image

and that some of the comments from https://github.com/UN-OCHA/common_design/pull/162#pullrequestreview-587992966 should be on this component instead.

left23:
Thanks @lazysoundsystem
This is tricky - I think I might have achieved what I was going for by adding:
```
div[class*="__title"] > *:not(a) {
  color: inherit;
  font-size: inherit;
  font-weight: inherit;
}
```
But let's see how this progresses.
The upside is it is unlikely that there will ever be a variety of markup like we have in the HTML __title div, which means we would be unlikely to see the different colours.
I think a sound expectation is to have the link colours apply when the titles are links.

PR #165, opened 2021-02-08
CD-226 cd-bullet-list

left23:
Review of the `cd-bullet-list` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-bullet-list for an example


lazysoundsystem (commented):
The changes I made need a bit more thought (there are now two `.cd-bullet-list li` declarations), but I wanted to differentiate this from the link list by not having all the items as links. 

I also tried with a different icon, which doesn't work very well as it's too small (and the wrong color).

left23:
@lazysoundsystem Oh ya, nice catch about the links not always being links. It even says so in the readme.

I've attempted to merge the `cd-226-design-system` branch  and resolve the conflicts, as you noticed I made some changes around RTL and icon markup in the main branch. I understand you are still working on this so I hope I didn't break anything.

I've also made the SVG icon selector more specific to the `arrow-right` when it flips.

If you refer to the README, it specifies the chevron icon (design decision) so I don't really think we have to cater for unknown other icons.
But generally, we should specify the arrow-right icon when flipping for RTL. There might be other occasions where this could be refined.

left23:
@lazysoundsystem I've made some changes based on your feedback. I needed this in order to work on updating the implementations. Feel free to still provide feedback.

lazysoundsystem (approved):
Sorry for the delay on this.

All looks good - feel free to discard my last commit, but if the styling only works for the chevron (which is a reasonable design decision) then we probably don't need the icon variant.

left23:
Thanks @lazysoundsystem I take your point about the SVG icon variant not being needed but I don't think removing it is necessary as it might still be something we develop either with other icons and some further development, or as was intended, as slightly different implementation technique.

PR #166, opened 2021-02-11
CD-226 cd-article

left23:
Review of the `cd-article` component

This is a strange one and maybe not needed, except that I wanted an article with a faux sidebar layout example on the demo page. The main layout rules come from the base theme and will likely stay there https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/sass/cd/cd-layout/_cd-layout.scss#L233 until we have cause to refactor all the layout rules.

Example of where it is used https://interagencystandingcommittee.org/the-inter-agency-standing-committee

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-article for an example


lazysoundsystem (approved):
Looks good - I tried with more content in the sidebar than the main section and it seems to handle that okay too.

In the .html file, there are a couple of leftovers that could go, the cd-byline stylesheet and the svg sprite.

PR #167, opened 2021-02-11
CD-226 cd-flow

left23:
Review of the `cd-flow` component.

This is like a utility so doesn't have its own demo. It can be seen on many components parent divs `[ cd-flow ]` https://feature.commondesign-unocha-org.ahconu.org/demo#cd-bullet-list

It's good to have a read of the CubeCSS documentation for general awareness, and as you can see, I'm still getting my head around it.


lazysoundsystem (approved):
In the section “But what about old browsers?” in https://24ways.org/2018/managing-flow-and-rhythm-with-css-custom-properties/ (written by the author of cubecss, I think) there's also a fallback `margin-top: 1em;` for browsers without custom properties.

I was looking for a different name for 'bumper' but I couldn't think of a better option and now I quite like it.

There's a commented out version for ltr bumping - I think it can be safely removed.

PR #168, opened 2021-02-11
CD-226 cd-filter

left23:
Review of the `cd-filter` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-filter for an example


lazysoundsystem (approved):
(Most of my comments for this got moved to the `cd-form` component.)

Without javascript, the form items are vertically stacked. With js, we don't want the cd-flow class.

I've changed the css counting for the 768px media query to use '4 and over'.

I got quite tied up in the css quantity queries. It seems children of cd-filter__form can either have the 'cd-filter__group' class or the 'block' class but can't be mixed. Would cd-filter-element (or something better named) make it less verbose. Then cd-export - or any other element, like submit or reset buttons (do we need those?) could be included in those. Let me know if a PR for that makes sense.

left23:
> (Most of my comments for this got moved to the `cd-form` component.)
> 
> Without javascript, the form items are vertically stacked. With js, we don't want the cd-flow class.

There is a rule to compensate for the addition of cd-flow on the first element when it's applied to the full-width `.cd-filter--horizontal`. We do want the cd-flow in most instances because it spaces out the other elements when they are in stacked columns (less than large desktop). When the filter elements are 2 columns, we have to make sure the first element also has a top margin so things are not misaligned. I think this is in place and working.
 
> I've changed the css counting for the 768px media query to use '4 and over'.

I'm not sure I like this change, but I see where you are coming from. This initially came from CERF https://cerf.un.org/what-we-do/allocation-summaries and one of the issues we had was needing to limit the amount of elements due to space. If we keep the 4 and over, the README is also no longer accurate "up to 5 elements".  We were specific here to prevent people asking us to "just add another element" because space issues make that hard. I consider this a design constraint.

> I got quite tied up in the css quantity queries. It seems children of cd-filter__form can either have the 'cd-filter__group' class or the 'block' class but can't be mixed. Would cd-filter-element (or something better named) make it less verbose. Then cd-export - or any other element, like submit or reset buttons (do we need those?) could be included in those. Let me know if a PR for that makes sense.

There are different layout rules for `.cd-export` and the others so they have specific selectors for this reason. The `.cd-export` aligns right, the others do not.
The `.block` are from Drupal blocks (when we add facet blocks in a sidebar for example) and the `.cd-fitler__group` is for when we have more control over the markup (like CERF). I think it is unlikely we will find ourselves in a position where we will mix both. I take your point that it breaks if the markup is mixed, but I would tend to wait for that to occur and adjust this then because I don't see it happening.

Let me know what you think. I'll hold off merging this.

lazysoundsystem:
Thanks Lisa - the issue with the flow spacing when they're horizontal was because I was using an old version on the demo site (sorry).

I only changed the separate counts for 4 or 5 to remove lines of CSS - perhaps just updating the comment, which I forgot to do, would be enough: "If four or five elements." in line 139.

And the suggestion of restricting the markup was also to make the CSS more concise, and I can see it's not an obvious trade-off to make the html more verbose to save on css.

How about:
```
  /* If five elements */                                                                                   
  .cd-filter--horizontal .cd-filter__form > *:nth-last-child(5):first-child,                                                      
  .cd-filter--horizontal .cd-filter__form > *:nth-last-child(5):first-child ~ * {                                                 
    flex: 1 0 20%;                                                                                         
    max-width: 20%;                                                                                        
  }             
```
It works for the demo and should cover the block variant too.

Aside from that (and let me know if a PR would help) it looks good to merge.

left23:
@lazysoundsystem Thanks. A more complete PR would be helpful because then I know it's been properly tested.
My single hesitancy at this moment to revisit this is the amount of testing I've done to date for all the variations. I know what is there works under the various scenarios it should. I appreciate refining the CSS to be more simple, but more important to me is that is works as expected.
If you would like to pursue this, please feel free. I think directly on this PR is fine. Otherwise maybe a TODO in the readme for future development.

lazysoundsystem:
I owe a big apology on this one. I've had the tab open in my browser for nearly three months now, waiting to make a PR of my suggestion, but now I look at it again I see it would only save very few lines.

So, adding approval. The next reviews will come quicker!

left23:
@lazysoundsystem Lol no probs. I thinks this warrants a refactor to simplify further when there is next the chance. I'm all for reducing lines of code but maybe more important to me is simplicity and clarity, and I agree, the selectors in this component can use some work. Leaving this open for now.

left23:
I came across a bug today traced to https://github.com/UN-OCHA/common_design/commit/b17d98fe9de1f599197059c1cb7dcf661cba1c56#diff-f521880bd24fbba8514175b9e3ee182db472306e1efb98e6903c254aa568f41f
So I've reinstated this as it is necessary for the sidebar to display in IE11.

PR #169, opened 2021-02-11
CD-226 cd-form

left23:
Review of the `cd-form` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-form for an example

*This one wasn't on the review list in the README, so I've added it.


lazysoundsystem (approved):
If we're not using `input[type='date']` should we have css styles for it? 

If labels aren't wrapped in a div, or separated with an `<hr>`, they don't get the right margin-top. there's probably a better way, but adding `display: block` for labels fixes that.

The -moz-placeholder rules should be safe to remove: https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder

Not sure what the `spinner-disabled` is doing - I see a couple of components have that with the double-bounce classes, but can't see any rules to do anything to them.

Might be good to add a `input[type='file']` and a `<textarea>` as well for demonstration (and in case they need any styling rules - they're both pretty common.

If the filter form goes across the page, I don't think we want/ need cd-max-width class on the `<p>` above it.

lazysoundsystem:
From cd-filter - another case where label would benefit from `display: block`:
image

left23:
Thanks @lazysoundsystem

> If we're not using `input[type='date']` should we have css styles for it?

No harm including it. This might not only be for Drupal.

> If labels aren't wrapped in a div, or separated with an `<hr>`, they don't get the right margin-top. there's probably a better way, but adding `display: block` for labels fixes that.

I take your point but I'm not going to go this far right now as I don't really have a particular use in mind for this component, other than to show the generic elements on the demo. Until I know more about specific intent, I predict any additions and corrections will be implementation specific, until we see patterns repeating.


> The -moz-placeholder rules should be safe to remove: https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder

Removed.

> Not sure what the `spinner-disabled` is doing - I see a couple of components have that with the double-bounce classes, but can't see any rules to do anything to them.

It currently does nothing. I have a note in the readme about developing it as a component. It is the visual indicator that the form filters are populating, which will be handy to have in some case. It's currently used on CERF forms, that's where it comes from. https://cerf.un.org/what-we-do/allocation-summaries

> Might be good to add a `input[type='file']` and a `<textarea>` as well for demonstration (and in case they need any styling rules - they're both pretty common.

Added type=file.
Textarea can be added when relevant.

> If the filter form goes across the page, I don't think we want/ need cd-max-width class on the `<p>` above it.

On the contrary, this is when it's needed most. We want to limit the length of any text beyond comfortable reading width. However, the actually text is copied from the cd-filter and so doesn't really serve a purpose here other than to show how any text could be added to a form element. I've removed it as I don't think it adds anything.

I'm tentative to develop this one too much because during implementations, we will need to make a lot of adjustments specific to Drupal markup and its forms and I don't know how to predict this. In time, we can give more back to this component as we find ourselves repeating common things on implementations.

PR #170, opened 2021-02-11
CD-226 cd-pagination

left23:
Review of the `cd-pagination` component

See https://feature.commondesign-unocha-org.ahconu.org/demo#cd-pagination for an example

### DIFF
--- a/components/README.md
+++ b/components/README.md
@@ -26,7 +26,7 @@ Add a check if you've reviewed the component and it passes the Criteria.
 | cd-hero             | :heavy_check_mark:  |                     |
 | cd-image-grid       |                     |                     |
 | cd-link-list        | :heavy_check_mark:  |                     |
-| cd-pagination       |                     |                     |
+| cd-pagination       |        :x:          |                     |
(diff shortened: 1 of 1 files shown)


lazysoundsystem (approved):
Added myself as a reviewer as per https://humanitarian.atlassian.net/browse/CD-249?focusedCommentId=121846

Looks good - Jonathan already mentioned `@supports` - I didn't notice it on the page (i.e. I think it doesn't harm usability) but having two chevrons for both 'next' and 'last' is confusing. I think the standard is '>' for the next page and '>>' to jump to the end.

left23:
Thanks @lazysoundsystem
I've change the chevrons to be single. However it's worth noting that the following is Drupal default when pagination is added to Views
Screenshot from 2021-03-04 13-28-07

FYI There is a template override in the base theme https://github.com/UN-OCHA/common_design/blob/cd-226-design-system/templates/navigation/pager.html.twig which controls this as well but it will be overridden by Drupal config.
@orakili suggested we replace the text character chevrons with SVG. I've added that as a TODO in the readme.

PR #173, opened 2021-02-16
CD-208 h1 page title

left23:
See [CD-208](https://humanitarian.atlassian.net/browse/CD-208) for background.

Summary:
1. remove h1 around logo in branding block
2. add page title block override with a comment about adding the `visually-hidden` class for the frontpage
3. add the relevant preprocess hooks from GHO and make changes to suit base theme and remove any irrelevant code.
4. refactor layout markup and rules to accommodate the page title being a child in the `main` and `article` elements, whereas before it was a sibling of `main`
5. Adjust the main Grid layout rule to accommodate the change in layout markup


sass/base/_typography.scss
```diff
@@ -131,9 +131,10 @@ h1 {
 }
 
 .page-title {
+  margin-top: 0;
```
left23, line 134: This was not done in v1 so the page title has a top margin, and will result in different vertical placement that will fail all VRTs unless subtheme has overriding rules.

sass/cd/cd-layout/_cd-layout.scss
```diff
@@ -35,15 +35,24 @@ body {
   }
 
   // Grid is used to fix the footer to the bottom of the screen, regardless of page min-height.
+  // Header height is fixed.
+  // Highlighted should contain Drupal messages and will be present even if empty.
```
left23, line 39: I checked a few implementations, GHO as well, and I can see the highlighted region div render despite no visible content. I think this is fine, and we can use this region to hold the Drupal messages - this might be standard Drupal.

This means a potential configuration change is also needed when updating the implementations, to move the status messages to the correct region.

sass/cd/cd-layout/_cd-layout.scss
```diff
@@ -35,15 +35,24 @@ body {
...
+  // Main should grow to fit the available space.
+  // Soft footer is optional.
```
left23, line 41: The grid layout works if the soft footer is absent but it assumes the other 4 elements are present. I think this is a safe assumption as a default.
On pages that might have a hero (or whatever full viewport element) as an additional sibling, this will need to be adjusted.

templates/content/node--full.html.twig
```diff
@@ -90,10 +89,16 @@
...
+    {% if not page %}
+      <h2{{ title_attributes }}>
+        <a href="{{ url }}" rel="bookmark">{{ label }}</a>
+      </h2>
+    {% else %}
+      <h1{{ title_attributes.addClass('page-title') }}>
+        {{ label }}
+      </h1>
+    {% endif %}
   {% endif %}
```
left23, line 102: This is an unlikely condition since it's the `node--full.html.twig` template but I wanted to have the logic represented in code somewhere, in case the template is copied for other node view mode templates.
I also didn't want to have a `node.html.twig` template because it seems too generic and thus unuseful, but I can introduce one if the conditional title heading logic is better suited there for clarity.

templates/content/page-title.html.twig
```diff
@@ -0,0 +1,32 @@
...
+ *   displayed after the main title tag that appears in the template.
+ * - is_front: Whether this the title of the homepage or not.
+ */
+#}
+
+{#
+  Add
+  is_front ? 'visually-hidden'
+  to classes array when there is no need for a visible h1 on the frontpage.
+#}
```
left23, line 21: For implementations, this will be copied into the subtheme if needed.

common_design.theme
```diff
@@ -238,3 +285,43 @@ function common_design_preprocess_menu(&$variables, $hook) {
...
+    $plugin_block = $block_manager->createInstance($id, []);
+
+    // Skip if there is no block matching the id.
+    if (empty($plugin_block)) {
+      return [];
+    }
+  }
+
+  return [];
+}
```
left23, line 327: @orakili Can you inspect this to double check only relevant code is included? I copied from SLT and GHO and removed what I thought wasn't needed.

templates/page.html.twig
```diff
@@ -76,33 +76,36 @@
...
-      {% if page.sidebar_first %}
-        {# We are using CSS pseudo class :empty to hide these aside elements if the region prints empty.
-          This means we cannot have any white space between tags (as below).
-          Attn: Having the twig debug enabled means the div is no longer empty.
-          See Drupal issues/953034 for description and current status.
-        #}
-        <aside class="cd-layout-sidebar-first" role="complementary">{{ page.sidebar_first }}</aside>
+      {% if page.page_title %}
+        {% block page_title %}
+          <div class="cd-page-title">
```
left23, line 85: @orakili I notice the wrapper div `cd-page-title` doesn't seem to get rendered. It's not important as there are no rules associated with it any more, but I'm curious as to why exactly.

left23:
I've added everyone as reviewers, mainly FYI. If you want to get into the details, please feel free, but not necessary.

left23:
To test this locally, the Page title block in Drupal should be moved into the content region, otherwise it will not appear on non-node pages.

common_design.theme
left23: I'll also add the access check from https://github.com/UN-OCHA/slt8-site/blob/develop/html/themes/custom/common_design_subtheme/common_design_subtheme.theme#L170

left23:
@orakili @rupl FYI I've deployed this to https://feature.commondesign-unocha-org.ahconu.org - when logged in, I should see the tabs below the Page title but I don't.
I'm also adding this to the node template override - https://github.com/UN-OCHA/gho-site/blob/develop/html/themes/custom/common_design_subtheme/templates/nodes/node--article--full.html.twig#L113
I assume it's a necessary piece.

PR #174, opened 2021-03-02
CD-193 jest tests

left23:
Set up of Jest in base theme with config and tests from https://github.com/UN-OCHA/reports-site/tree/dev/_tests and https://jestjs.io/docs/en/puppeteer

**Steps**:

1. Check out PR
2. In `common_design/` run `nvm use` and `npm i` and `npm i puppeteer`
3. Run `npm run test` and/or `npm run e2e`


left23:
When I run 'npm run test' I get issues around the env variable.
Screenshot from 2021-03-02 15-47-15

I'm not sure where the env files come into things - How is the env being detected?
Do I need to change the env file in [index.js](https://github.com/UN-OCHA/common_design/pull/174/files#diff-95aaa17b462eb4ff6932af54b48f86f650ba150e814c0d5bf14e62bb37169928R4) to use locally? 

@rupl Could you help or comment on the above? I get the same result if I run `npm run e2e` or `npm run test`


Notes below to myself. I'm making progress.

If I remove the env import: 
```
const timeout = 15000;
const baseUrl = 'https://commondesign.test';
jest.setTimeout(timeout);

describe('AppBar', () => {
  beforeAll(async () => {
    await page.goto(`https://reports.unocha.org`);
  });
```
I get different errors, maybe puppeteer-related:
Screenshot from 2021-03-02 15-54-45

As I'm reading all this, I realise I should be running `npm run e2e` because of this:
https://github.com/UN-OCHA/common_design/pull/174/files#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1R1
which I understood to be the same as https://jestjs.io/docs/en/puppeteer#use-jest-puppeteer-preset

So now I'm getting missing Puppeteer errors.

```
npm WARN jest-puppeteer@4.4.0 requires a peer of puppeteer@>= 1.5.0 < 3 but none is installed. You must install peer dependencies yourself.
```

Oh.

`npm i puppeteer`

To be continued.

_tests/env/index.js
```diff
@@ -0,0 +1,6 @@
+//
+// Change which environment gets loaded by updating where the import points to.
+//
+import env from './production.js';
```
rupl, line 4: > I'm not sure where the env files come into things - How is the env being detected?
> Do I need to change the env file in index.js to use locally?

Yes. This is the line hardcoded to prod. There's no mechanism for detecting environments; you change this line to run against local.

rupl:
In your screenshot it says `page is not defined` and that means Puppeteer is not being loaded. The test files expect a few puppeteer-related globals to always exist (`browser`/`page`), much like how we expect `window`/`document` to exist in a browser.

If it doesn't exist, we need to figure out why it's not getting automatically loaded. I'm happy to pair with you tomorrow after our morning meetings.

rupl:
FYI `page` holds just about everything you'd want if you are simulating user actions like navigation/clicking/typing so we will have to figure that out before you can go much further. Here's the official reference: https://pptr.dev/#?product=Puppeteer&version=v8.0.0&show=api-class-page

left23:
> In your screenshot it says `page is not defined` and that means Puppeteer is not being loaded. The test files expect a few puppeteer-related globals to always exist (`browser`/`page`), much like how we expect `window`/`document` to exist in a browser.
> 
> If it doesn't exist, we need to figure out why it's not getting automatically loaded. I'm happy to pair with you tomorrow after our morning meetings.

Yep, I updated the PR to include puppeteer install. It was a separate step I needed to do, and I would have known this but I was ignoring the npm errors (there are some that are unrelated). And assuming puppeteer came with jest and jest-puppeteer.

I'm getting results now when I bypass the env import. 

I changed the index.js to point to local.js but I'm still getting the first error about the env variable. It's not stopping me for the moment.

rupl:
@left23 for the `import env` problem, what version of node is this being run inside? wherever you write "npm run e2e" what is the result of `node -v`?

left23:
> @left23 for the `import env` problem, what version of node is this being run inside? wherever you write "npm run e2e" what is the result of `node -v`?

v10.18.1 which is currently what is in the .nvmrc for the base theme.
I'm aware this has to be updated. I was hoping to tackle all this when I removed gulp etc but if it's a blocker for the tests, I can separate these tasks.
If it's likely the node version causing issues, leave it with me and I'll revisit when I get a chance.

rupl:
honestly just try bumping to `v12.21.0` and it will probably unblock you without breaking too much

left23:
I've upgraded node
```
node -v
v12.21.0
```

So far, I still get the same issue trying to include the .env file.

PR #175, opened 2021-03-03
[CD-268] Component review

orakili:
Ticket: CD-268

@left23 Sorry the https://github.com/UN-OCHA/common_design/pull/110 PR is too heavy for my computer and I was unable to add inline comments for the components so I've compiled all my notes here. Let me know if you want that in a different format.

---

Common Design review
====================

Generalities
------------

### Sass

I think using the `css variables` for the values of the sass variables would
help consistency, maintenance and overriding in the subtheme.

So instead of `$cd-blue--dark: #025995;` have `$cd-blue--dark: var(--cd-blue--dark)`.
This could probably apply to most sass variables, including breakpoints etc.

### Testing helpers

Some components like the `cd-grid` have testing helpers adding a background
color for example. Is it something that will be removed before the first
release of the v2?

Components
----------

### CD-FORM

This component should probably be revisited later to help make filling forms
more consistent across websites which could help the content squad for example.

#### CSS

`select` seems to be the only element that always spans the full width unless
marked as inline.

`cd-form--inline`: the `display: flex` makes the height of the elements (ex: input)
different that other *non-inline* elements of the same type making forms less
consistent.

`margin-inline-end` used for inline labels doesn't seem to be supported in IE 11.

`fieldset ul`: this seems really generic and could impact cases where it's
legitimate to display the bullet points in a fieldset (ex: field instructions),
it would probably be better to have a class for lists of checkboxes/radios.
I even wonder if it's really semantic to have checkboxes/radios in `ul`.

`outline-offset` is not supported in IE.

The blue `outline` on focus only applies to input but probably should apply to other
for elements like `select` and `textarea` for consistency.

**Search input cancel button**

Normalize.css (line 366) hides the search cancel button, which arguably is a
useful feature. As the README of that component states that we're not trying to
have the same behavior across browsers, I think it could be good to restore this
cancel button.

```css
/**
 * Remove inner padding and search cancel button in Safari and Chrome on OS X.
 * Safari (but not Chrome) clips the cancel button when the search input has
 * padding (and `textfield` appearance).
 */

input[type="search"]::-webkit-search-cancel-button,
input[type="search"]::-webkit-search-decoration {
  -webkit-appearance: none;
}
```

#### Examples

Some of the elements that are styled in this component are not represented in
the twig template:

- input types aside of "text"
- textarea

---

### CD-GRID

Is the naming with a single dash on purpose: `cd-grid-2-col` rather than with
a double dash to indicate they are variants like `cd-grid--2-col`?

#### CSS

The `grid-3-col` has an extra `max-width` on desktop that the other grids don't
have:

```css
@media (min-width: 1024px) {
  ...

  .cd-grid-3-col > *  {
    flex: 0 0 32%;
    max-width: 32%;
  }

  ...
}
```

The `clear` rule line 67 and the IE hack at the end could be removed as there is
no notion of floated elements anymore.

I also wonder if the `cd-grid--grow` should not be replaced by a class added to
the grid element that should use the full width rather than , if that should not
be a class that could be applies to any element of the grid rather than just the
last one:

```css
.cd-grid .cd-grid-item--grow {
  flex: 1 0 100%;
  grid-column: -1/1;
  max-width: 100%;
  max-width: unset;
}
```

That would notably allow to have a header and footer spanning the full width
inside a grid layout.

---

### CD-HERO

The hero image is not limited anymore by the 1400px `max-width` of the v1 and
can result in very large images on large screens. That may have quite an impact
on performances and as they are not bound in width anymore, difficult to
generate appropriate responsive image styles for large screens.

#### CSS

In `cd-hero--style-one.css`, all the rules aside of the `background: linear-gradient(...)`
could probably be removed as they are for browsers we don't support anymore.

When `object-fit: cover` is not supported, the hero images appear deformed due
to the fixed height on the image itself. It might be better to wrap the image
in a container that would have the fixed height and `overflow: hidden`. For
example always have `` wrapped in a `<picture>`.

---

### CD-IMAGE-GRID

What is the use case for this component with images limited to 64px?

Could we provide variants for different sizes?

---

### CD-LINK-LIST

There is no visual feedback when the chrevron is a link and it's difficult
to understand that one needs to click on it especially when there are other
links like in the second example.

Maybe the link in the footer should the entire width and height of the footer
to increase the clickable area in addition to the visual feedback.

---

### CD-PAGINATION

Looks good. I would suggest a few modifications to the template for
people who use it as reference:

- Replace `<span class="visually-hidden">Current page</span>` with
  `<span class="visually-hidden">Current page, page 1</span>` (added current
  page number to the label) to improve accessibility because otherwise
  `current page` is meaningless especially if someone reached the current page
  from a link found somewhere and has no notion of which page it is.
- Add a `aria-current="page"` attibute to the `<a>` of the current page to
  also improve accessibility.
- Replace the arrow characters (ex: `<`) with the corresponding svg icons.

#### CSS

The css uses `@supports (display: grid)` but it should be `@supports (display: flex)`
as grids are not used for this component.

---

### CD-READ-MORE

Looks good.

---

### CD-SEARCH

I would argue that the template has too many drupal related classes that
makes obfuscate the core elements and classes of the component.

The svg icon is also missing the classing `aria-hidden` etc.

There is also a `aria-labelledby=cd-search-btn` is incorrect and should refer
to the `h2` to which an `id` should be added.

---

### CD-SEARCH-INLINE

Same comments as the `cd-search`.

The `page--demo.html.twig` wraps the component in a `<div>` with a relative
position which causes the component to look like it's behaing differently than
the `cd-search` on mobile as it doesn't span to the border. Is it on purpose?

---

### CD-SELECT-ALLY

Skipping as cannot test on the common design site.

---

### CD-STYLED-LIST

Looks quite similar to the `cd-link-list` which could be a subset of this
styled list?

In the demo on mobile and in RTL, the footer doesn't change and its content stays aligned
right.

#### CSS

Should the last block with the `.cd-layout-sidebar-first` be in that
component?

---

### CD-TABLE

Some issues with the components already being discussed in https://humanitarian.atlassian.net/browse/CD-219

#### CSS

The rules for the `caption` are too opiniated (disply below the table) and classes
should probably be used to define the behavior, ex: `cd-table--bottom-caption`.

#### JS

There are lots of issues with the javascript:

- The function should be wrapped in a `iife` to avoid polluting the global scope.
- It applies only to the first responsive table: `document.querySelector('.cd-table--responsive');`
- `Array.from` is not needed and that would remove the need for a polyfill.
- For the label put in `data.content`, the innerHTML should not be used otherwise
  it will copy more than the text. For example if you have `<th><strong>header</strong></th>` then
  the entire `<strong>header</strong>` will be saved in the `data-content` attribute
  but will be displayed as if displaying the encoded html: `&lt;strong&gt;Type&lt;/strong&gt;`.
- The function doesn't handle dynamically added responsive tables and could use
  a dom observer or something like that to apply the logic to newly added tables.

---

### CD-TABS

Looks good. A few questions regarding the css rules below.

#### CSS

Why the following code for the clearfix:

```css
ul.tabs::after {
  content: '.';
  display: block;
  height: 0;
  clear: both;
  visibility: hidden;
}
```

Rather than the classic:

```css
ul.tabs::after {
  content: '';
  display: table;
  clear: both;
}
```

Is the `.` important?

Are the `!important` necessary? Any example where drupal is troublesome and
those are required?

---

### CD-TEASER

Looks good.

### CSS

IE doesn't support `unset` so the below block may not behave as itended:

```css
.cd-teaser__container {
  padding-left: unset;
}
```

---

### CD-TITLE-LIST

Looks good. Just wondering if it should not use the `cd-bumper` like the
`cd-link-list` component to separate the link from the chevron.

---

### CD-TOC

Looks good.

---

### CD-TYPOGRAPHY

Not sure about this component. I would tend to agree with the proposal to
split that into separate components for the blockquotes, definition lists etc.

Also, several of the typography rules, like for the headings are in
`sass/base/_typography.css` so I wonder if either the ones in the sass file
should be moved to that component or the other way around.

#### CSS

The `article p` setting the max-width, seems opiniated and limited (what about
`article ul` for example?). I would remove that and just keep the  `.cd-max-width`.
There is also a similar rules in the `cd-article` component.

The blockquote color should probably use a css variable.

The rules for the `dl` are assuming a single definition per term. They don't
work well for `dt` followed by several `dd` which is a legitimate use case (ex:
list of theme for an article). I would propose to remove the rules for the `dl`
etc. and classes instead with variants like `cd-definition-list--inline-single`.

That could be a separate component with different variants.

---

### CD-UTILITIES

Looks ok. I wonder if it should not be merged with `cd-flow` as both define
utility classes.


left23 (approved):
I've reacted to what I could in the short term, added backlog items or TODOs for things that can be improved as needed.
Thanks for the detailed feedback.

PR #176, opened 2021-03-10
[CD-273] Refactor page title for nodes

orakili:
Ticket: CD-273

This refactors a bit the way to display the page title on node pages.

*Note: there are some unrelated changes which are just coding standards fixes*

## Setting

There is now a setting on `/admin/appearance/settings/common_design` to select for which node view mode, the `page title` block should be pre-rendered and displayed inside the node `<article>` instead of using the default placement of the page title block.

Screen Shot 2021-03-10 at 18 44 24

By default, the behavior is enabled for the node "full" view mode.

## Functions

The function that retrieves the block render array `common_design_get_block_render_array()` now stores the list of regions in which the block was supposed to be rendered so that it can be removed from the page via `common_design_remove_rendered_blocks()` called in the `common_design_preprocess_page()`.

The handling for nodes is done in the `common_design_preprocess_node()`. The logic could be applied to other pages (ex: views) by using the same functions in another hook.

## Templates

The `node--full.html.tpl` file was modified, removing the wrapping `<h1>` as it's already set in the `page-title.html.tpl` template. This was causing the double `<h1>`.

**Note:** When enabling the behavior for another view mode, the implementers should copy the part of the `node--full.html.tpl` template with the logic for the title.

## Attributes

The `title_attributes` property for the page title block has been enabled (see `common_design_theme_registry_alter()`). It was mentioned in the `page-title.html.tpl` but it had no effect.

## Patch

There is no need anymore for this patch: https://github.com/UN-OCHA/common-design-site/blob/7039f4ed19216bb48a11bbc4367bb85c142c46ac/composer.patches.json#L5

## Instructions

I think some instructions should be added somewhere in the repo about the setting and the necessity to clone the `node--full.html.tpl` template.

### DIFF
--- a/common_design.theme
+++ b/common_design.theme
@@ -104,9 +104,17 @@ function common_design_preprocess_page(&$variables) {
 function common_design_preprocess_node(&$variables) {
   $view_mode = $variables['view_mode'] ?? '';
 
+  // Get the list of node view modes for which to use the pre-rendered page
+  // title instead of the normal page title block. This is an associative array
+  // keyed by view modes and with either the view mode as value when enabled
+  // or 0 otherwise. If not set, enable the behavior for the full view-mode.
+  $view_modes = theme_get_setting('common_design_node_title');
+  if (!isset($view_modes)) {
+    $view_modes = ['full' => 'full'];
+  }
+
   // Prepare the title and local tasks so we have better control over where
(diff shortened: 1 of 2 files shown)


left23 (approved):
Perfect, thanks @orakili

PR #177, opened 2021-03-10
CD-193 jest tests

left23:
To run tests in the common_design base theme directory:

Install packages `nvm use` `npm i`

If you have a local stack set up, you can run the tests on that by replacing the [config file import](https://github.com/UN-OCHA/common_design/pull/177/files#diff-8bf3db05a2f0060670267ce31c1c2528a256527e1f09f7ce02dfa7e239e8c38dR4) in `_tests/_env/index.js` to `local.js`

I run tests against the [demo](https://demo.commondesign-unocha-org.ahconu.org/) or the [feature](https://feature.commondesign-unocha-org.ahconu.org/) site, both which use http authentication with secrets, so I've added a separate config locally, and added that file to the [gitignore](https://github.com/UN-OCHA/common_design/pull/177/files#diff-e26b0f45a94405cd348443f862434a02a1712063bb3396217ee4e3cc3b325761R1)

Run all tests `npm run e2e`

I would expect the OCHA Services test to fails as explained on [CD-193](https://humanitarian.atlassian.net/browse/CD-193?focusedCommentId=123947) but all others should pass.

Screenshot from 2021-04-07 11-21-20


left23:
@rupl General review please, at your leisure.
I'm interested in whatever feedback you have, and also are the test names clear, and if you think any tests are not present but should be.

_tests/FooterMenu.test.js
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+  it('should contain specific links in the Footer menu', async () => {
+    const footerMenuItems = [
+      'Privacy',
+      'Terms of use'
+    ];
+    const footerMenuItemsHref = [
+      'https://www.un.org/en/sections/about-website/privacy-notice',
+      'https://www.un.org/en/sections/about-website/terms-use'
```
rupl, line 18:
```suggestion
      'https://www.un.org/en/sections/about-website/privacy-notice/',
      'https://www.un.org/en/sections/about-website/terms-use/'
```

This test was failing until I added trailing slashes:

CD-193-FooterMenu

rupl (approved):
Looking great to me! Nice work 💯

_tests/FooterMenu.test.js
rupl: Whoops, meant to include that it was the Feature site where the test failed.

left23: @rupl I'm gonna opt to change the URLs in the footer on the Feature site, rather than change the test. I assume the URL with the trailing slash was a copy-and-paste error and not intentional or necessary.
Also making a note to myself here as I noticed the UN site has redirects in place and the URLs could be https://www.un.org/en/about-us/privacy-notice and https://www.un.org/en/about-us/terms-of-use instead.

PR #179, opened 2021-03-12
[OPS-7333] Encode ';' in google fonts URLs to fix issue with CSS aggregation

orakili:
Ticket: OPS-7333

Durpal 8/9 CSS aggregation breaks on the `;` inside the google fonts URLS (https://www.drupal.org/project/drupal/issues/2936067).

This PR encodes the `;` so that we don't need to apply a patch to the sites using the common design theme.


attiks (approved):
Slightly related, another solution is to host the fonts on he site, will make the site faster as well

orakili:
> Slightly related, another solution is to host the fonts on he site, will make the site faster as well

Yep, there is a ticket for that: https://humanitarian.atlassian.net/browse/CD-244

rupl (approved):
This seems like a nice short-term fix 👍

PR #183, opened 2021-03-18
[CD-276] Use drupal libraries to add google fonts

orakili:
Ticket: CD-276

This PR proposes to use drupal libraries to load google fonts so that they can easily be enabled/disabled in a subtheme for example. To be discussed with the team.


rupl:
ooh! purely in theory I like where this is headed. I wonder if we can we also generate `:root` blocks from the libraries file so that enabling the drupal library, or using as a dependency automatically injects the custom prop? 🤔

README.md
```diff
@@ -100,10 +100,49 @@ Run `npm run sass:build` to compile production-ready CSS.
...
+
+Fonts for other languages like Chinese can be added in the `common_design_subtheme.libraries.yml`:
+
+```yaml
+# Chinese.
+# @see https://brand.unocha.org/d/xEPytAUjC3sH/visual-identity#/basics/fonts-1/chinese
+fonts-chinese:
+  css:
+    theme:
+      //fonts.googleapis.com/css2?family=Noto+Sans+CJK+SC:wght@400;700&display=swap: { type: external, minified: true }
```
left23, line 125: @orakili How did you find the Font URL for Noto CJK SC?

orakili: @left23 The use of Noto Sans CJK SC for Chinese comes from the brand site: https://brand.unocha.org/d/xEPytAUjC3sH/visual-identity#/basics/fonts-1/chinese which has a link to https://www.google.com/get/noto/#sans-hans. I just "built" the URL based on the `Noto+Sans` by adding the `CJK+SC` which apparently is wrong. The correct URL seems to be `//fonts.googleapis.com/css2?family=Noto+Sans+SC:wght@400;700&display=swap` (no CJK).

left23: Ok, that's what I arrived at too, but I don't fully understand the implications of dropping the CJK - I get the impression using only SC might result in less/no support for additional languages like Korean and Japanese (?). Also I don't really know how to test this with any authority. 
AFAIK there are no implementations with Chinese as a requirement (bar WHD but it uses another font chosen by the designer) so I think it's safe to mark this as "needs work / needs testing" and include what we have now as is.

orakili: The fonts for Japanese, Korean, Chinese etc. are actually separate in https://www.google.com/get/noto/ and in the API: https://fonts.google.com/?query=noto+sans. The only difference is the naming with the extra "CJK": "Noto Sans CJK JP" for Japanese vs "Noto Sans JP", etc.

So for Japanese that would be "Noto Sans JP", for Korean "Noto Sans KR". For Chinese there are 2 fonts: SC (simplified Chinese, used in China mainland) and TC (traditional Chinese, used in Taiwan for example). To be sorrow we would probably need to check the second part of the language code to see which font to use (zh-CN for SC and zh-TW for TC). 

Chinese is one of the [official UN languages](https://www.un.org/en/our-work/official-languages) so I think we should include it. My understanding is that in that context, Chinese refers to the simplified version so SC.

PR #184, opened 2021-03-19
[CD-276] Move google fonts import out of :root

orakili:
Ticket: CD-276

Quick fix that moves the `@import` of the google fonts out of the `:root` because they were not loaded when there.


left23 (approved):
Approving, thanks @orakili

PR #191, opened 2021-03-30
CD-269 replace gulp

left23:
Remove Gulp and associated packages and config.

Add npm scripts for:

1. Sass: Compile, lint, watch and postcss. Run `npm run sass:build` for final build (and on the server) and `npm run sass:watch` for CSS with source maps.
2. JS lint
3. SVG sprite build
4. e2e tests from #177 

```
  sass:compile
    sass sass/styles.scss css/styles.css --style=compressed --no-source-map
  sass:compile-dev
    sass --embed-sources sass/styles.scss css/styles.css
  sass:watch
    npm run sass:lint && npm run sass:compile-dev -- --watch
  sass:lint
    stylelint '{sass,components}/**/*.{css,scss}'
  sass:postcss
    postcss css/styles.css --use autoprefixer -b 'last 3 versions' -r
  sass:build
    npm run sass:lint && npm run sass:compile && npm run sass:postcss
  js:lint
    eslint js/*.js
  svg:sprite
    svg-sprite --svg-xmldecl=false --svg-doctype=false --svg-rootattrs=_config/svg-sprite.config.json --log=info --symbol --symbol-prefix='%s' --symbol-example=true --shape-id-generator='cd-icon--' --symbol-dest=img/icons --symbol-sprite=cd-icons-sprite.svg img/icons/*.svg
  e2e
    jest _tests/*.test.js
```

svg-sprite brings with it dependency updates that I have reinstated. 
https://github.com/svg-sprite/svg-sprite/issues/446

I opted to drop [sass-lint](https://github.com/sasstools/sass-lint) as it is deprecated and use [stylelint](https://stylelint.io/user-guide/usage/cli) instead. Stylelint config is included in Drupal core so I've used that and [extended the config](https://github.com/sasstools/sass-lint) in the base theme as needed, including [setting the severity from error to warning](https://github.com/UN-OCHA/common_design/pull/191/files#diff-528119d81e5a10738ae49f5e20d333ebc37cea63657eb9069281693e6dae0ad3R11)

Includes https://github.com/UN-OCHA/common_design/pull/195 and https://github.com/UN-OCHA/common_design/pull/198


index.js
```diff
@@ -0,0 +1,24 @@
+import sass from "sass";
+import { promisify } from "util";
+import { writeFile } from "fs";
+const sassRenderPromise = promisify(sass.render);
+const writeFilePromise = promisify(writeFile);
```
rupl, line 5: I believe you have imported the proper promise-based `writeFile` already. I don't think you have to promisify it.

index.js
```diff
@@ -0,0 +1,24 @@
+import sass from "sass";
+import { promisify } from "util";
+import { writeFile } from "fs";
+const sassRenderPromise = promisify(sass.render);
```
rupl, line 4: the Sass docs actually claim that using the sync render function is 2x as fast (although I'm sure we're splitting hairs for a build process) — so maybe we don't need to promisify it either.

https://github.com/sass/dart-sass#javascript-api

rupl:
I tested the latest branch just now and the SVG seems to be working. This was my repo diff once I ran the command, which seems encouraging if none of the underlying sprites were updated in the process:

```diff
-                       <p>Generated at Wed, 31 Mar 2021 15:40:36 GMT by <a href="https://github.com/jkphl/svg-sprite" target="_blank">svg-sprite</a>.</p>
+                       <p>Generated at Thu, 01 Apr 2021 07:20:06 GMT by <a href="https://github.com/jkphl/svg-sprite" target="_blank">svg-sprite</a>.</p>
```

left23:
@rupl Can you have another look at this at your leisure.
Overall, I'm happy with how it works but I can't seem to get the Sass linting to happen during the watch task. It lints initially, but doesn't lint changes after watch is run initially.
This isn't s show stopper but I'm curious to see if you can spot why this is or if you have suggestions.

.stylelintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": [
+    "../../../core/.stylelintrc.json",
```
rupl, line 3: I happen to have a free-standing `common_design` repo and this caused a fatal error when spinning up. I think we can ignore that because the theme will almost always sit inside a Drupal installation.

Just wanted to note it in case we can make a mention in the docs or something, since up until now it was working independent of a drupal installation.

rupl: when I say free-standing I mean it's in a place such as `~/src/common_design` outside of a drupal theme directory.

rupl: I think in general it's smart to be re-using Drupal config so aside from mentioning in docs that the error could pop up when used outside Drupal, no change necessary to the code.

README.md
```diff
@@ -125,13 +124,13 @@ Using `this` works for most functions except ones which are assigned to event li
...
+## Task management
 
-This project uses [Gulp 4](https://github.com/gulpjs/gulp#whats-new-in-40)
+This project uses some Node packages for Sass compilation, watching and linting, JS linting and SVG icon sprite generation.
 
-To see a task listing, run the following command:
+See [scripts in package.json](https://github.com/UN-OCHA/common_design/blob/main/package.json#L9)
 
-`gulp --tasks`
+Eg: `npm run sass:watch`
```
rupl, line 133:
```suggestion
To get a list of commands, do `npm run` and it will output all possible options:
```

.stylelintrc.json
rupl: ok, I moved into a drupal installation and got the same error, but the file WAS indeed found in the expected relative path. I ended up resolving it by running `npm i` inside Drupal core itself. That fixed it for me.

rupl:
In terms of running the linter on file-save, I wouldn't expect that to happen unless we also pass that tool some sort of "watch" option (and it indeed supports such an option). It's not jumping out at me when I browse the CLI flags: https://stylelint.io/user-guide/usage/cli#options

If you're ok adding another dependency, I created #194 to implement `npm-watch` and drive both lint+compile off that tool.

.stylelintrc.json
left23: Great, thanks for pointing out the non-Drupal perspective. I'll add these details to the readme.

left23: I'm always torn about these decisions. Using the Drupal one makes it handy, but if we are catering for non-Drupal as well, having config in the base theme makes it more likely to be useful elsewhere and reduces the documentation (arguably).

There's an [eslint config file](https://github.com/UN-OCHA/common_design/blob/main/.eslintrc) in the base theme, and when I went looking at how we arrived at the decision, I got back to [CD-184](https://humanitarian.atlassian.net/browse/CD-184) and https://github.com/UN-OCHA/common_design/pull/23

@orakili Any recent thoughts on this? For the Drupal 8/9 version of RW, would you opt to keep the eslint config in the theme, or use Drupal core eslint config?

left23:
@rupl @lazysoundsystem @orakili 
I'm contemplating removing this githook https://github.com/UN-OCHA/common_design/blob/main/.githooks/pre-commit/pre-commit.sh as it was not always useful (it regenerates the CSS even if the changes are not to styles) and I couldn't always get it working in the subthemes, so I got into the habit of committing the generated CSS. Any thoughts on keeping this, or extending it to inlcude more useful things?

Are there other implementations that have useful githooks already?

lazysoundsystem:
I'm not aware of any - but I wasn't aware of this one either. 

Anyway, it's good to compile the css before pushing commits to check if the change does what it's supposed to.

left23:
> Anyway, it's good to compile the css before pushing commits to check if the change does what it's supposed to.

@lazysoundsystem I agree. I'm not sure if I understand your full meaning here though. Typically, if there is work done on the styles, the dev will either be running the watch or run the specific compile command. If we are working locally, we'd do this and then check our work. Are you saying the githook of regenerating the CSS is useful in case the CSS was not compiled by the dev?

rupl:
> Any thoughts on keeping [the git-hooks], or extending it to inlcude more useful things?

I've always tried to like them, but the pre-commit and post-save hooks actually get in my way quite often. I basically appended `--no-verify` to every commit I ever made in GHO, because there was some problem finding the binary and I never had the time to sit and debug the environment.

When linting actually becomes a frequent problem on a project, I first opt for pre-push hooks. Develop how you like, but lint before sharing your work. Nice middle ground IMO.

lazysoundsystem:
@left23 sorry for not being clear. 

> Are you saying the githook of regenerating the CSS is useful in case the CSS was not compiled by the dev?

No, I was saying the githook shouldn't be necessary because the css should already be compiled.

left23:
@rupl @orakili Can I draw your attention to this commit https://github.com/UN-OCHA/common_design/pull/191/commits/2860b35a0b3db81c8cc1ba57b4868360c657a482
There's more info on [CD-269](https://humanitarian.atlassian.net/browse/CD-269)
I've more work to do on this PR. When I'm done, I'll request a review from you both.

rupl:
I think you've chosen a precise workaround. Great work hunting this bug down!

left23:
@rupl @lazysoundsystem @orakili I've added reviewers as FYI and/or if you would like to test or input.

I think this is an adequate first pass and might need refining as it gets used in upcoming projects and/or as I prepare to deploy the v3 updates of all implementations.

There is still work to do updating the subtheme to match v3, so I expect more changes and clarification to the README in next iterations, to include the distinction between which tests are for the sub theme specifically.

.stylelintrc.json
rupl: To follow-up: coding standards and linting are theoretically most effective when our whole team uses one standard across multiple projects. So given that Drupal is our most frequently-used tool, IMO it's smart to use it as a standard.

I'd also move to implement the result of this PR into HID, for instance. I might have to copy some Drupal files, or tweak how we include the linting config, but ultimately if that means anyone on our team can pop into HID and make a change without much friction, we have improved the maintainability of HID by adopting the standard provided by Drupal.

rupl (approved):
I've been testing this incrementally several steps along the way, so I don't have any surprise opinions to pull out at this stage 😁 

It's all working for me, including the E2E using your docs on how to exclude the Subtheme from the run 💯 Great job!

lazysoundsystem:
I added this to my common-design-site setup and had to make a couple of changes before I could get `npm run sass:build` to complete.

I got:
`Error: Could not find "stylelint-config-standard". Do you need a `configBasedir`?`
which I fixed with:
`npm install stylelint-config-standard`

and then similar ones for `stylelint-no-browser-hacks` and `stylelint-order`.

I found these in the drupal stylelintrc.json, so I assume all is good (and that those should be added to package.json).

Also, just noting that I got this warning.
`npm WARN jest-puppeteer@4.4.0 requires a peer of puppeteer@>= 1.5.0 < 3 but none is installed. You must install peer dependencies yourself.`

Otherwise, all looks good!

rupl:
> jest-puppeteer@4.4.0 requires a peer of puppeteer@>= 1.5.0 < 3

I checked and the [latest branch](https://github.com/smooth-code/jest-puppeteer/blob/a7b569351fc6be97f6e24b6f2967d47ddd6f4329/package.json) specifies puppeteer@8.0.0 in `devDependencies`, which is the current release. I think this can be ignored, and will generally correct itself over time.

left23:
@lazysoundsystem Thanks. I forgot to add the stylelint + Drupal aspect to the README. I think running `npm install` in the Drupal project root should install get the necessary packages, including the https://github.com/drupal/core/blob/8.9.x/.stylelintrc.json#L2 but I need to test this more pointedly.

We discuss it here a bit https://github.com/UN-OCHA/common_design/pull/191#pullrequestreview-628937340

I'm still experimenting with how extends work https://stylelint.io/user-guide/configure#extends

I think the goal is to have as little as possible in the base theme stylelint config, and instead rely on Drupal core's https://github.com/drupal/core/blob/8.9.x/.stylelintrc.json

lazysoundsystem (approved):
Tested and it all works well (though the linting has a lot to say!).

orakili (approved):
Tested on drupal 9 (reliefweb) and worked well after running `npm insall` in the drupal core directory.

I think for now it's ok to just add that instruction (or any other way) in the README but probably worth to see if we can retrieve the styling standards file as part of the install in the common design theme.

PR #194, opened 2021-04-06
CD-269: use npm-watch to lint+compile on each filesave

rupl:
Adds a lint-on-save function to the ongoing PR #191

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,7392 @@
 {
   "name": "common-design",
   "version": "3.0.5",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "common-design",
+      "version": "3.0.5",
+      "license": "GPL-2.0",
+      "dependencies": {
+        "autoprefixer": "^10.2.5",
+        "sass": "^1.32.8"
+      },
+      "devDependencies": {
+        "eslint": "^7.23.0",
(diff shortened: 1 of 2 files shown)


left23:
@rupl Thanks.
This matches the current Gulp set up, where the linting runs on watch (as part of `gulp dev`). For me, this is handy because I can fix the linting issues as I go. The alternative (if we don't merge this PR) would be to get the list of warnings when running `sass:build` either manually or as part of the pre-commit git hook or via travis, which is also reasonable.

@orakili @lazysoundsystem Did you use the `gulp` or `gulp dev` task (which has the watcher) in the past when you worked on the sub themes? Or did you compile after you made some changes using `gulp build`?

lazysoundsystem:
Not often, no - generally when I've made changes to the styling I've used `gulp sass`.

orakili:
I also tend to simply use `gulp sass` (linting is set to run when saving in my editor).

rupl:
I'm also ok with running the linter on Travis and reacting to it when desired, instead of every filesave. The command can always be run manually. I only put the PR together because of our discussion in #191 and I assumed it was your wish to restore this bit of the workflow.

Happy to close this if our team won't get much regular use out of it 👍

left23:
@rupl Thanks for preparing this. It's a nice option for the future if I really feel the linting is missing from the watch process. I should have asked people's opinion beforehand. In order not to add another dependency, I think we can close this without merging.

PR #195, opened 2021-04-07
CD-269 Sass and CSS linting with stylelint

left23:
Branched from https://github.com/UN-OCHA/common_design/pull/191


rupl:
Just FYI you can change this PR's base branch to be the #191 branch so the diff shrinks, and when it eventually gets merged, this PR's base will automatically change back to `main`

PR #196, opened 2021-04-09
CD-285: Adds base theme property to theme .info.yml as that is now re…

berliner:
…quired


left23:
Thanks @berliner, merging.

PR #198, opened 2021-04-13
CD-284 refactor linting

left23:
Merge after https://github.com/UN-OCHA/common_design/pull/195

1. Remove following rules from stylelint config: 

 ```
"order/properties-order": null
"order/order": null,
"string-quotes": null,
```
and fix using `--fix`

2. Add to script to fix and expand README, reduce CONTRIB

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -5,12 +5,10 @@
   ],
   "rules": {
     "no-descending-specificity": null,
-    "function-calc-no-unspaced-operator": null,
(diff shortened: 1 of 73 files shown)


rupl:
I spotted a couple very small things I consider to be made slightly more confusing with the linting rules, but nothing so drastic as to make an exception. Is there VRT to support this branch?

left23:
Yep, I'm merging this and I can deploy a feature branch to run the VRTs on the Feature site.

PR #205, opened 2021-04-20
Fix/ops 7412 subtheme

orakili:
Ticket: OPS-7412

Fix issues with the subtheme being inside the theme:

1. Make sure Drupal doesn't use the subtheme templates inside the common_design folder.
2. Rename the `common_design_subtheme.info.yml` file so that the subtheme inside the theme is not detected as it's meant to be copied elsewhere.


left23 (approved):
Thanks, merging into the base branch to test.

PR #209, opened 2021-04-23
CD-288: explicitly upgrade font requests to HTTPS

rupl:
# CD-288

Upgrades fonts to always use HTTPS. See ticket for some links and references for why.


left23 (approved):
Thanks @rupl

rupl:
Whoops having reviewed this one more time, it seems I forgot to compile the sass and update the CSS. Just a heads up for when it deploys and we test it.

PR #213, opened 2021-04-29
[IASC-683][CD-283] Fix the cacheability of the page title and local tasks block

orakili:
Tickets: IASC-683, CD-283

This fixes the issue with the double page titles and local tasks.

The problem was happening when the rendered node was cached and the page was rebuilt then the `hook_preprocess_node().` was not called and the list of rendered blocks was not populated so the function called in `hook_preprocess_page()` couldn't remove the page title and local tasks from the "page title" region.

This solves the problem by storing the rendered block info (id, region) as a cache tag which then can be retrieved whether the rendered node is cached or not.

This also adds more appropriate cache metadata to the page title and local tasks blocks.

### DIFF
--- a/common_design.theme
+++ b/common_design.theme
@@ -5,6 +5,7 @@
  * Template overrides, preprocess, and hooks for the OCHA Common Design theme.
  */
 
+use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Form\FormStateInterface;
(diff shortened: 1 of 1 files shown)


attiks (approved):
Nice!

left23 (approved):
Approving and merging, thanks @orakili

PR #214, opened 2021-04-30
[CD-283] Fix double page titles

orakili:
Ticket: CD-283

This fixes the previous fix to fix the double page titles...

It uses a different approach as we cannot rely on the cache metadata to determine if the page title and local tasks blocks were rendered (ex: IASC which renders the node in a block making the cache metadata for the node unavailable in the hook_preprocess_page).

Instead it retrieves the view mode for the node in the `hook_preprocess_page()` and if it's one of the selected view modes in the theme settings, then it hides the page title and local tasks blocks from their page region.

### Tests

Tested the fix against IASC, CD site, AR, SLT ([needs a patch in subtheme](https://github.com/UN-OCHA/slt8-site/pull/26)), ODSG and GHO.

### Consideration about rendering/hiding other blocks

Like previously, the same approach as what is done for the page title and local tasks blocks can be used for other blocks (`common_design_get_block_render_array()` + `common_design_hide_rendered_blocks_from_page()`).

The previous (problematic) method(s) had the advantage of being a bit more generic. For example only a call to `common_design_get_block_render_array()` was needed to render a  block and then the `hook_preprocess_page()` would hide the corresponding block automatically

The new method is more trustable but requires an explicit call to `common_design_hide_rendered_blocks_from_page()` to hide the block rendered via `common_design_get_block_render_array()`.

See https://github.com/UN-OCHA/slt8-site/blob/9e1e4cb216aeb93e0a54c3ecb1277d1db054b1e5/html/themes/custom/common_design_subtheme/common_design_subtheme.theme#L93-L137 for an example.

### DIFF
--- a/common_design.theme
+++ b/common_design.theme
@@ -5,6 +5,7 @@
  * Template overrides, preprocess, and hooks for the OCHA Common Design theme.
  */
 
+use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Cache\CacheableMetadata;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\user\Entity\User;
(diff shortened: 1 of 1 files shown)


left23 (approved):
Thanks @orakili Approving and merging.

PR #221, opened 2021-05-12
OPS-7440 reinstate e2e debug statements

left23:


_tests/_env/index.js
```diff
@@ -14,4 +14,6 @@ const environments = {
   },
 };
 
+console.log('⚠️ E2E testing', environments[process.env.NODE_ENV].baseUrl);
+
```
rupl, line 18: instead of doing this in the environments file, which duplicates the info a bunch of times, maybe we should put it in one of the tests which always runs.

The other potential problem with having this particular line of code is that if `process.env.NODE_ENV` isn't set, it will create an error. The next line has a conditional which guards against that situation.

left23: @rupl I don't think we can predict which test will always run since these will be managed in the sub theme in most implementations. In this case, I don't see issues with duplicate info. It makes more sense to me for this to be part of the config.

As per the second point, I don't understand what you are getting at. If `process.env.NODE_ENV` isn't set this will be a problem beyond the console statement, won't it? What are you suggesting?

rupl: Yes, it will go beyond the console.log and cause the test to fail.

If `process.env.NODE_ENV` is set to empty string, OR a value that doesn't exist in the config, then the value of `environments[process.env.NODE_ENV]` will be `undefined` and then we get a ReferenceError because we're trying to log the property `baseUrl` of `undefined` 👇 

Screen Shot 2021-05-12 at 16 01 31

To protect against that situation, we can use a similar pattern as the next line to ensure the object exists before trying to display the property value:

```js
console.log('⚠️ E2E testing', environments[process.env.NODE_ENV] && environments[process.env.NODE_ENV].baseUrl);
```

left23: I'm not sure how to test this. If I run `export NODE_ENV=whatever` locally, and then run `npm run e2e` I get the same result as your screenshot above - where baseUrl is undefined

rupl: I guess my code suggestion wasn't correct.  I will try a couple things and make a PR to this branch.

PR #222, opened 2021-05-13
[OPS-7440] output baseUrl safely

rupl:
I gave some bad, untested code suggestions in the other PR https://github.com/UN-OCHA/common_design/pull/221. This is fully tested. Set `NODE_ENV` to any value at all (including empty string) and this should work.


left23 (approved):
Looks good @rupl, thanks.
Approving and merging.

PR #230, opened 2021-06-01
[CD-226] cd-grid review

lazysoundsystem:
I might have garbled the todo in the README - please check it makes sense!

The last-child approach seems to make more sense when there's just one item in the last row. Where there's more than one, you get an 'extra line':
image
Is that as planned?


left23:
Thanks @lazysoundsystem 
I'm going to leave this one open for a while. @rupl might be using this for another implementation and I might soon as well, so this is an opportunity to decide whether it is robust enough, or whether another grid system would suit better.

left23:
@lazysoundsystem Can you elaborate on what behaviour you would expect from your comment in the description?

left23 (approved):

PR #237, opened 2021-06-01
[CD-226] approve component

lazysoundsystem:
Just noting that the .html template, unlike the others, is a full html document with head and body sections (and 'cd-toc--float-left' class while .twig template has 'cd-toc--float-right'.


left23:
Thanks, noted. 
Some of the html files have surrounding container divs to better represent what the component will be like in situ. This was one of those cases. Without the layout divs, the article spans 100% viewport width instead of being within the bounds of the content area with the max-width. The twig files don't need the layout divs because the components are output via the `page--demo.html.twig` where those divs are present.

left23 (approved):

PR #238, opened 2021-06-01
[CD-226] approve component

lazysoundsystem:
Looks good, just noting two 'extra' layers of html which I don't think are useful:
```
  <main role="main" id="main-content" class="cd-container">
    <div class="cd-layout-content">
```


left23 (approved):

left23:
Noted. Same as comment on other PR https://github.com/UN-OCHA/common_design/pull/237#issuecomment-853957963

PR #240, opened 2021-06-08
CD-269: add theme hook suggestions for any Views Display

rupl:
# CD-269

New feature: Twig template suggestions for any Views Display being rendered.

While implementing CD on a project, I had a particularly protracted battle with Drupal trying to get some Twig templates to be recognized in regards to a specific Views Display. In the end, the code to enable the behavior was quite short and I'd like to add it to Common Design as a more dev-friendly default.

## Before

It either didn't render any suggestions at all, simply reporting that it pulled the CD `views-view.html.twig` or it listed the one option with no alternatives:

```html
<!-- FILE NAME SUGGESTIONS:
   x views-view.html.twig
-->
```

## After

This is a real example of a View with machine key `features` and Display machine key `embed_latest_4`:

```html
<!-- FILE NAME SUGGESTIONS:
   * views-view--features--embed-latest-4.html.twig
   * views-view--features.html.twig
   x views-view.html.twig
-->
```

### DIFF
--- a/common_design.theme
+++ b/common_design.theme
@@ -357,6 +357,22 @@ function common_design_theme_suggestions_page_alter(array &$suggestions, array $
 }
 
 /**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ *
+ * Provides more granularity when Views displays are getting rendered. The two
+ * suggestions will allow the following templates to be picked up:
+ *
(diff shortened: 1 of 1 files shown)


left23 (approved):
Great @rupl Handy addition.

PR #244, opened 2021-06-15
CD-299: apply cursor:pointer to all CD Buttons

rupl:
# CD-299

Applies `cursor: pointer;` to all CD Buttons.


left23:
Merge it when you're ready.

PR #247, opened 2021-06-17
CD-210 site manifest in sub theme

left23:
As an alternative to the PWA module.
Also added some details to the README.


left23:
@rupl Can you have a quick read and confirm it's clear, or let me know if I should add more details.

common_design_subtheme/site.webmanifest
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "Common Design Demo",
+    "short_name": "CD",
```
rupl, line 3:
```suggestion
    "name": "YOUR WEBSITE NAME",
    "short_name": "WEBSITE",
```

IMO the values here look a little too "finished," and I'd rather put default values that are plainly in need of attention.

common_design_subtheme/site.webmanifest
```diff
@@ -0,0 +1,20 @@
...
+        {
+            "src": "img/android-chrome-512x512.png",
+            "sizes": "512x512",
+            "type": "image/png"
+        }
+    ],
+    "theme_color": "#1f69b3",
+    "background_color": "#ffffff",
+    "display": "standalone",
+    "start_url": "."
```
rupl, line 19:
```suggestion
    "start_url": "/"
```

When I look at [MDN docs for start_url](https://developer.mozilla.org/en-US/docs/Web/Manifest/start_url#relative_url) it says:

> If the URL is relative, the manifest URL is used as the base URL to resolve it.

So if I'm understanding correctly, a value of `.` would end up pointing to `/sites/themes/custom/common_design_subtheme/`

README.md
```diff
@@ -203,3 +203,10 @@ npm run e2e -- -t '^(?!.*Subtheme).*$OCHAServicesDropdown'
 # All tests except blocks including 'Subtheme'.
 npm run e2e -- -t '^(?!.*Subtheme).*$'
 ```
+
+## Progressive Web App
+
+[web.brand.unocha.org](https://web.brand.unocha.org) website uses the Drupal 
+[PWA module](https://www.drupal.org/project/pwa)
+
+There is a `site.webmanifest` file available in the sub theme as an alternative to the module. The file should be adjusted per implementation, and added using a `<link>` element in the `<head>`, with additional configuration needed. See https://developer.mozilla.org/en-US/docs/Web/Manifest for implementation details.
```
rupl, line 212:
```suggestion
There is a `site.webmanifest` file available in the sub theme as an alternative to the module, if all you want is the Web Manifest, and you don't want offline capability. The manifest file should be adjusted per implementation, and added using a `<link>` element in the `<head>`, with additional configuration needed. See https://developer.mozilla.org/en-US/docs/Web/Manifest for implementation details.
```

rupl (approved):
Overall looks good, and I do think it's a simpler solution for people who only want a manifest. 👍

PR #251, opened 2021-07-02
CD-297 Remove base/_grid.scss partial, improve cd-image-grid and Views Grid format

left23:
- delete grid partial, remove import statement, regen css
- Add CSS custom property to `cd-image-grid` for grid item size, with default value
- Improve Views Grid format so `cd-grid` works for the Horizontal alignment (most common) and the Vertical alignment keeps working as intended (See component README for details)
Selection_025


lazysoundsystem (approved):
Looks good - I made some picky whitespace and line length changes in a new commit (hope that's okay) and tested it locally on the common-design-site.

Unrelated to these changes, but relevant to the grid - the '-grow' variant works well when there's one element in the last line, but can look odd when there's more than one, e.g.
image
and
image
or
image

left23:
@lazysoundsystem Thanks. You've mentioned this before and I was curious to find out how you think it should work https://github.com/UN-OCHA/common_design/pull/230#issuecomment-858504970

lazysoundsystem:
Sorry, in taking the screenshots i thought I might have been repeating myself, and I missed your question on the other ticket. 

My expectation for the grow variant is that it's basically 'justify' - so each row is the same length. Does that make sense or is this intended for grids where the last item is a special item, (a button or something), that should 'underline' the grid?

If there's no remainder - e.g 6 items in a 3 column grid or 12 in a 4-column grid (the first two examples in the screenshot above) - I think the last item should slot into place in the grid. 
```
|_|  |_|  |_|
|_|  |_|  |_|
```

If there's a single remainder - e.g. 7 for 3-column grid or 13 for 4-column grid - then the 'extra' item should stretch to 100%.
```
|_|  |_|  |_|
|_|  |_|  |_|
|___________|
```

If there are e.g. 3 remaining items in a 4-column grid, I'd imagine grow would make them 25%,25%,50% rather than 25%,25%,line break, 100%. (The third example in the screenshot above).
```
|_|  |_|  |_|  |_|
|_|  |_|  |_|  |_|
|_|  |_|  |______|
```

left23:
@lazysoundsystem Thanks for the feedback. My intention for adding the `grow` variant was more to provide a way to clear the other grid items, like a footer would i.e. explicitly separating the item from the rest of the elements in the grid.
See Homepage https://assessments.hpc.tools as an example.

Another example we can use is IASC (I'm replacing the custom grid with cd-grid) currently on the dev site https://dev.interagencystandingcommittee-org.ahconu.org/

Here is how a the `cd-grid--4-col` works currently:
Selection_036

If the last item expanded to fill the available space (if I understand your suggestion correctly) we would end up with:
Selection_037

Your suggestion makes sense as an expectation, but I think it complicates things, and I'm not sure we have a use case at this point. In the example of IASC, the teaser would also need more work to react to multiple wider-width situations, as would any other component we might predict would be placed in a grid.

I've added only one small comment to the cd-grid component so maybe this is part of the problem. I could do better explaining the intent.
https://github.com/UN-OCHA/common_design/tree/main/components/cd-grid

```
// Last item to span full width.
.cd-grid--grow
```

I can only reiterate that if we want flexible grid layouts, we should probably find a more robust grid system than what the component provides.

lazysoundsystem:
Thanks @left23 that makes perfect sense - and it's a good lesson for me when I get involved with the changelog documentation: the comment in the README explains exactly what it does, but I managed to overlook it.

components/cd-image-grid/cd-image-grid.css
```diff
@@ -1,6 +1,11 @@
+:root {
+  --cd-grid--item-size: 4rem;
+}
+
 .cd-image-grid > * {
   display: inline-block;
-  padding: 0.25rem;
+  /* Padding is a 16th of the grid item size */
+  padding: calc(var(--cd-grid--item-size, 0.25rem) / 16);
```
orakili, line 8:
```suggestion
  padding: calc(var(--cd-grid--item-size, 4rem) / 16);
```

components/cd-grid/README.md
```diff
@@ -24,7 +24,20 @@ This would allow a header and footer span the full width inside a grid layout.
...
 
 ## Caveats
-The selectors must be added to the parent div
+The selectors must be added to the parent div.
+
+### Using with Drupal Views Grid format
+The Views UI allows configuration for the number of columns, and alignment
+(horizontal or vertical). The `cd-grid` component is used **only when the
+Horizontal alignment is selected**. The checkbox for **Automatic width
+should be unchecked** and **maximum number of columns is 4**. See the template
```
orakili, line 33: Just wondering if we should not add the `cd-grid` only when those 2 conditions are met (in addition to the `horizontal` alignement).

components/cd-image-grid/cd-image-grid.css
```diff
@@ -1,6 +1,11 @@
+:root {
+  --cd-grid--item-size: 4rem;
```
orakili, line 2: I think this should be named `--cd-grid-image--item-size` to avoid confusion (and potential collision in the future) with the `cd-grid` component.

orakili (commented):
Looks good but I would rename the scale variable in the `cd-image-grid` component to `--cd-grid-image--item-size`.

templates/views/views-view-grid.html.twig
```diff
@@ -26,16 +26,13 @@
...
-{{ attach_library('common_design/cd-grid') }}
-
 {%
   set classes = [
     'views-view-grid',
     options.alignment,
     'cols-' ~ options.columns,
     'clearfix',
-    'cd-grid cd-grid--' ~ options.columns ~ '-col',
+    options.alignment == 'horizontal' and automatic_width == false ? 'cd-grid cd-grid--' ~ options.columns ~ '-col',
```
orakili, line 35: @left23 I think it should be `options.automatic_width` here and below, according to the [preprocess function](https://api.drupal.org/api/drupal/core%21modules%21views%21views.theme.inc/function/template_preprocess_views_view_grid/8.9.x).

left23: Thanks, I've updated that.

PR #248, opened 2021-06-17
CD-301 Details about Annual Release schedule in README

left23:
- Details about the release schedule in the readme
- Changelog file with last patch release as first entry and current commit status under Unreleased section

PR #253, opened 2021-07-12
CD-276 multilingual webfonts

left23:
Branches from https://github.com/UN-OCHA/common_design/pull/183

I've added a partial for the Roboto font, so it is compiled in the styles.css

The other fonts (including Roboto Condensed and Slab, which are currently not used in any design elements) are defined as libraries to be included as needed.

The only disadvantage is that since these fonts are not included in the base theme by default, we cannot demonstrate the multilanguage fonts on https://web.brand.unocha.org without special treatment in the sub theme, which I do not want to do because it would mean including the libraries in the sub theme as default and that's not what we want to encourage.

I've added Chinese from the README example on this PR https://github.com/UN-OCHA/common_design/pull/183

I've updated the Arabic library to load Noto Kufi and use for headings as per brand guidelines.

I've updated the README to describe the font management, and corrected some whitespace issues.

Additional, and not related to this PR, I've removed reference in sub theme README to remove reference to renaming the sub theme as per comment https://github.com/UN-OCHA/common_design/pull/248#discussion_r663835028


sass/base/_base.scss
```diff
@@ -35,13 +35,13 @@ body {
   }
 }
 
-html[lang="ar"] body {
+html[lang="ar"] body,
+html[lang="ru"] body {
   font-family: $cd-font--noto;
```
orakili, line 40: Noto Sans doesn't handle Arabic, so that would result in the system fonts being used for Arabic.

sass/cd/_cd-variables.scss
```diff
@@ -75,16 +75,17 @@
...
-
   // Roboto
   // https://brand.unocha.org/d/xEPytAUjC3sH/visual-identity#/basics/fonts/advanced-users
   --cd-font--roboto: "Roboto", helvetica, arial, sans-serif;
   --cd-font--roboto-condensed: "Roboto Condensed", helvetica, arial, sans-serif;
   --cd-font--roboto-slab: "Roboto Slab", helvetica, arial, sans-serif;
+
+  // Multilingual
+  // https://brand.unocha.org/d/xEPytAUjC3sH/visual-identity#/basics/fonts/multi-language-typefaces-1
+  --cd-font--noto: "Noto Sans", sans-serif;
```
orakili, line 86: I would argue that the variable should be named `--cd-font--noto-sans` to better reflect the real font name (smae above with `noto-kufi-arabic` and `noto-sans-sc`, notably to be more explicit and differenciate between Noto Sans and Noto Serif for example.

sass/base/_typography.scss
```diff
@@ -112,16 +84,28 @@ html[lang="ar"] {
...
   h4,
   h5,
   h6 {
     font-family: $cd-font--noto;
+    font-weight: 400;
+  }
+}
+
+// @TODO ZH might need special attention.
+html[lang="zh-hans"] {
```
orakili, line 100: in the `_cd-resets.css` it's still `lang="zh"` and it's not using the correct Chinese font I think: https://github.com/UN-OCHA/common_design/blob/5f70fdc64f4bdb34214b3ef3c667eb90691c949a/sass/cd/_cd-resets.scss#L44

common_design.libraries.yml
```diff
@@ -1,15 +1,28 @@
...
       //fonts.googleapis.com/css2?family=Roboto+Slab:wght@400;700&display=swap: { type: external, minified: true }
 
-# Arabic.
+# Multilingual
 # @see https://brand.unocha.org/d/xEPytAUjC3sH/visual-identity#/basics/fonts/multi-language-typefaces-1
+# Arabic.
 fonts-arabic:
+  css:
+    theme:
+      //fonts.googleapis.com/css2?family=Noto+Sans:wght@400;700&display=swap: { type: external, minified: true }
```
orakili, line 15: Noto Sans doesn't handle Arabic: https://fonts.google.com/?preview.text=%D8%A7%D9%84%D8%AD%D8%A8%20%D8%B3%D9%85%D8%A7%D8%A1%20%D9%84%D8%A7%20%D8%AA%D9%85%D8%B7%D8%B1%20%D8%BA%D9%8A%D8%B1%20%D8%A7%D9%84%D8%A3%D8%AD%D9%84%D8%A7%D9%85.&preview.text_type=custom&query=noto

orakili (changes requested):
Looks good but there are issues with Arabic and the footer in Chinese. 

I expanded on this work with theme settings to add the fonts: https://github.com/UN-OCHA/common_design/pull/255

left23:
I noticed as I was setting up translation example locally that the current selectors are not adequate
```
html[lang="ar"] body {}
```
This works when the global language is set via the language switcher.
For Nodes set to specific languages, the `lang` attribute is on the article tag only.

rupl:
> `html[lang="ar"]` works when the global language is set via the language switcher.

Would it work to omit the `html` part of the selector? I often apply RTL fixes using just `[dir="rtl"]`. For example:

```css
.my-selector {
  margin-left: 1rem;
}
[dir="rtl"] .my-selector {
  margin-right: 1rem;
}
```

I'd have to setup a test page to see what happens when they nest, i.e `html[lang="en"] article[lang="ru"]`

left23:
@rupl I'll test the nested example. I've got some local content set up already.
From what I can see, when I switch to a node whose language I've set on the node form (Arabic for example), the global language switches as well. But if I switch to another language using the global switcher, while on that node, the global language switches but the node language remains as arabic. Which is all great and what should happen. Noting it here because it takes a lot of content set up to test properly.

I expect the selector will work as you suggest. Thanks.

left23:
@orakili Thanks, I think I've covered all your comments in the latest commits.
You mention an issue with the footer in Chinese. Can you be more specific?

orakili (changes requested):
@left23 Thanks. I added code suggestions to change `$cd-font--noto-sc` to `$cd-font--noto-sans-sc` to be consistent with the rest, and one for a Chinese selector in`resets.scss` one chinese.

orakili (approved):

PR #255, opened 2021-07-16
[CD-276] Add theme settings to enable multilingual fonts

orakili:
Ticket: CD-276

This PR extends https://github.com/UN-OCHA/common_design/pull/253 by adding some theme settings (ex: `/admin/appearance/settings/common_design_subtheme`) to allow site builders to enable multilingual fonts:

Screen Shot 2021-07-16 at 14 12 00

The list of available fonts is retrieved from the base theme and sub theme (when enabled) `libraries.yml` files (the `fonts-*` libraries).

So it's possible to add additional multilingual fonts (ex: Japanese in the screenshot) by adding a library in the subtheme libraries.yml file and then enabling it in the settings.

```yaml
fonts-japanese:
  label: Japanese
  css:
    theme:
      css/fonts-japanese.css: {}
```

### Changes

1. The multiingual fonts rules and variables have been removed from the sass files (only Roboto is there).
2. The libraries don't load the fonts directly but reference `css` files that contain the `@import`, `@font-face` etc. and the necessary rules to apply the fonts and define the relevant css variables.

The `@import` and rules to apply the font-family are regrouped in a fonts css files (ex: `css/fonts-arabic.css`).

For sites not using Drupal, one can simply add the css file in addition to the `css/styles.css` one.

An advantage of this approach, is that everything related to a font is regrouped in one css file. It's also possible to mix`@import` and other fonts like the `css/fonts-arabic.css` example in the subtheme which is using Noto Kufi from google fonts and Dubai, present in the subtheme.

**Note**: the `css/fonts-arabic.css` in the **base theme** only contains Noto Kufi Arabic (like on GHO). The `css/fonts-arabic.css` in the **sub theme** uses Dubai for the body (with files and font-face extracted from the DSR site).


orakili:
@left23 The settings introduced in this PR are available without the subtheme (at `/admin/appearance/settings/common_design`):

Screen Shot 2021-07-16 at 14 30 19

So it's possible to feature the multilingual styling in the Demo site by enabling the fonts and adding content in the corresponding languages.

left23:
@orakili This looks good, but let's hold off until we get confirmation about the Arabic font. I thought we'd agreed not to include font files in the repo. I think for now we can use the googlefont Noto Kufi Arabic for everything, like what was done in GHO instead of including font files.

orakili:
@left23 The Dubai fonts in the subtheme is just to illustrate what can be done with this approach (mixing `@import` and `@font-face` with self hosted fonts etc.). Happy to remove it and move that to the README.

rupl:
@orakili I'm trying to clear out our queue of CD PRs and I'm wondering if this one is still worth reviewing and bringing to the finish line. In the PR body you mentioned it's a follow-up to a merged PR so my gut says yes, but I'll let you confirm before I go rebasing from `main` to `develop` and re-prepping it for merging.

left23:
We never got confirmation about the Arabic.
The font spec here https://brand.unocha.org/d/xEPytAUjC3sH/visual-identity#/basics/fonts-1 is in line with https://github.com/UN-OCHA/common_design/pull/255/files#diff-95956bed7584614cada408eae8a9428f92efd184413e0966646e1d2cfa6d3510

@orakili Are there implications for the existing sub themes we should take note of?

common_design_subtheme/common_design_subtheme.libraries.yml
```diff
@@ -3,6 +3,15 @@ global-styling:
...
 
+# Arabic font using Dubai.
+# @todo This of the sake of the demo and could be merged with the base theme one
+# or removed.
+fonts-arabic:
+  label: Arabic
+  css:
+    theme:
+      css/fonts-arabic.css: {}
+
```
rupl, line 14: This libraries definition seems like it affects subtheme implementations. Perhaps we can leave it here commented out like some of the other examples.

rupl:
I'm reassigning to myself for the work of rebasing against `develop` which has drifted significantly (especially since `_cd-variables.scss` no longer exists and that's my doing).

After having a look today, I still think it's worth merging in. It removes yet another piece of the Sass pie while retaining the ability to be dynamic via theme settings.

rupl:
My same testing comments from #322:

I tested this branch on my local CDS and it did add the sub-theme font Library when I copied the example `common_design_subtheme.libraries.yml` definition into my active sub-theme:

CD-276-subtheme-font-library

After saving the settings as shown in the picture (no sub-theme library enabled) I did see the fonts applied to the public pages. I didn't bother taking a "before" screenshot but it was indeed using Helvetica beforehand and now the recognizable Dubai font is in place, verified in my devtools computed styles:

CD-276-fonts-applied

PR #262, opened 2021-08-20
[CD-327] Fix page title logic

orakili:
Ticket: CD-237

As discovered in https://humanitarian.atlassian.net/browse/IASC-716, there may be some occasions where a node is rendered in a view mode selected to use the page title block, outside of the node page, for example when indexing a rendered node with Search API.

Another problem with the current logic is that the node title and local tasks don't appear in the node edit page when using the CD theme for the edit forms.

This PR refines the logic to limit the use of the page title to node canonical and preview pages.

This is compatible with the current logic and so updating any site using the CD theme should be seamless.

**Test**

1. Update any site using the CD with the code from this branch.
2. Check that "full" is selected in the CD settings (should be by default).
3. Visit a node page and check that the node title is indeed using the page title template (enable twig debug if not already) and that the title is inside the `<article>` tag.
4. Edit the node and click preview and check the same as above.

Testing that the page title block is not used in other contexts is more complicated. This can be done for example, by rendering a node in full view mode via an entity reference field:

1. Create a temporary entity reference field for nodes on a node type.
2. In manage display of the node type, select "rendered entity" for the field and the "full" view mode.
3. Create a node and add a another node to the field
4. Save or preview and confirm that the referenced node is using the `node-full.html.twig` template but that the title is wrapped in a `<h2>` and not using the page title template.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Fixed
 - Dependency updates for xmldom
+- Fix page title logic to limit the use of the page title to node canonical and preview pages
 
 ---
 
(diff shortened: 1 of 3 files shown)


left23 (approved):
Thanks @orakili I tested both scenarios and everything works as described.

left23:
I've added the item to the changelog. @orakili Feel free to merge when ready. I think the other reviewers are as FYI

PR #296, opened 2022-02-04
test(ci): add a Travis CI config file

rupl:
testing to see if Travis picks this up automatically

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,5 @@ language: node_js
 node_js:
   - 12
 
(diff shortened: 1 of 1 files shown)


rupl:
ok I see now. we extend the Drupal core stylelint config and it doesn't exist without a far more involved CI preparation step that involves an entire Drupal site. bummer.

> Error: Could not find "../../../core/.stylelintrc.json". Do you need a `configBasedir`?

PR #322, opened 2022-04-26
[CD-276] merge develop into Jonathan's work

rupl:
Trying to make sure I didn't screw up the merge with `origin/develop`


rupl:
Nice, ok. It took a couple tries to be sure that I managed the merge conflicts correctly, but in the end both branches were deleting a lot of the same content, just in different ways.

I tested this branch on my local CDS and it did add the sub-theme font Library when I copied the example `common_design_subtheme.libraries.yml` definition into my active sub-theme:

CD-276-subtheme-font-library

After saving the settings as shown in the picture (no sub-theme library enabled) I did see the fonts applied to the public pages. I didn't bother taking a "before" screenshot but it was indeed using Helvetica beforehand and now the recognizable Dubai font is in place, verified in my devtools computed styles:

CD-276-fonts-applied

PR #329, opened 2022-05-24
CD-397 Improve styles when Javascript is disabled

left23:
## Types of changes

## Description
This PR improves the display of the search and refines the styles for the active menu items **when Javascript is disabled**.

## Motivation and Context

The search is too wide on mobile, it makes the search button hard to access.

Some styles relevant to the menu dropdowns are still applied despite the dropdown not being created. Hiding separators improves the visual clarity when the first level parent and second level children are displays.

The active menu styles are applied to the the first(parent) and second level active item but since Javascript is disabled, the parent item not longer functions as the dropdown button but is instead an anchor element. We don't want to emphasize this because the anchors lead nowhere, so the styles are adjusted to only display the second level item as active.

1. Disabled Javascript in the browser and notice how https://web.brand.unocha.org the menus (also when active) and the search input display on different viewport sizes. 
2. Check out this branch locally and disable Javascript. Notice how the elements are improved

The search width now defaults to the input size.

Selection_002

The search width expands as the menu items increase so this might look different depending on the amount of menu items and whether second level menus exist.

Only 2 first level menu items:
Selection_003

Multiple menu items and 2 second levels menus:
Selection_004

Active second level menu item:
Selection_005

Active menu items styles remain the same on mobile as before:
Selection_008

The parent item still gets the hover style because it's a link. I'm not sure what the approach here should be.
Selection_007

## Impact
No negative impact expected. Issues may arrive if sub themes have competing styles but that is unlikely.

## PR Checklist


orakili:
> [orakili](https://github.com/orakili) closed this in https://github.com/UN-OCHA/rwint9-site/pull/345[5 days ago](https://github.com/UN-OCHA/common_design/pull/329#event-6682086745)

No idea why merging the RW9 PR closed this CD PR... Sorry about that.

rupl:
It seems to think that "fix" meant the other PR fixes this one:

Screen Shot 2022-07-11 at 10 52 12

rupl (changes requested):
Nice improvement! I think it's far better to have reachable buttons than not, even if the layout is a little chunky. The merge conflicts seem to be due to the CSS Vars work, so I've reassigned myself to take it from here.

rupl (approved):
The conflicts weren't so bad after all!

PR #350, opened 2022-11-30
CD-231 Language Switcher

left23:
## Types of changes

## Description
As the history of CD-231 describes, this PR changes the display of the Language Switcher to display in full when space permits.

It also refactors the Language Switcher where relevant. For example, replacing `[dir="rtl"]` rules with `inline-start`, removing hacks for browsers we no longer support, and removing the `cd-user-menu__dropdown` which used the extends rules in  `sass/cd/cd-header/_cd-user-menu.scss`. The last item results in some duplication because the language switcher shares some aspects like the dropdown, the toggle button, but its markup is different to the other 2 blocks (both Drupal menu blocks).

There are also additional RTL rules that set positions for the elements in the Global Header. These can be reduced and replaced with Logical properties where possible.

The pseudo-elements that used to create pipes are all removed. In their place we use Flexbox and a linear-gradient to create the exact same visual effect, with fewer conditionals, RTL rules, and overrides in other components.

## Motivation and Context
Better aligning the design and behaviour to the UN.org Multilingual standard.

## Expected behavior
On mobile, the Language Switcher should display in a dropdown
On > mobile, it will display in full.

Selection_073
Selection_072
Selection_074
Selection_075

## Actual behavior
Regardless of breakpoint, the Language Switcher displays in a dropdown.

Since this requires changes to the base theme and the configuration, multiple steps are require to set this up locally.

- Check out this branch and clear the cache
- Follow the config changes or checkout the [common-design-site]( https://github.com/UN-OCHA/common-design-site/) branch `CD-231-global-header` and import the config

- Test with User Menu and or Additional Menus or Blocks in the Drupal `Header Top` region
- Test with JS disabled
- Test on mobile viewport
- Test with other languages, including RTL
- Test with a sub theme override (this is part of https://github.com/UN-OCHA/common-design-site/pull/225)

*Note: the active language only receives the `is-active` class for anon users on my local set up. 

## Impact
The default dropdown breakpoint is desktop but this might not be suitable for all sites, depending on what other elements (if any) are in the Global Header. In these cases, there are rules to copy into the sub theme to set the Language Switcher to display as a dropdown at another breakpoint.

There are also site configuration changes needed, see https://github.com/UN-OCHA/common-design-site/pull/225
This might vary per implementation.

## PR Checklist


sass/cd/cd-header/_cd-language-switcher.scss
```diff
@@ -6,123 +6,142 @@
...
-  margin-left: 12px;
-  padding: 0;
-
-  li {
-    display: inline-block;
-    padding: 0 12px;
+// Display the full Language Switcher on > mobile.
+.cd-language-switcher {
+  @include mobile {
+    --dropdown: false;
```
left23, line 132: I've set the default breakpoint for when the Language Switcher changes to a dropdown as mobile ( < 576 ). This needs to be tested. Maybe < 768px is more realistic. This can be tested on a few sites

sass/cd/cd-header/_cd-language-switcher.scss
```diff
@@ -6,123 +6,203 @@
...
+  &[dir="rtl"] .cd-global-header__language-switcher + .cd-global-header__user-menu::before {
+    right: -1px;
+    left: auto;
+  }
+}
+
+// Adjust breakpoints for when the Language Switcher displays as a dropdown.
+// This depends on other elements present in the .cd-global-header__actions.
+//
+// Copy the following rules into the sub theme and override breakpoint values.
```
left23, line 143: This corresponds with https://github.com/UN-OCHA/common-design-site/pull/225 where there is a component added to the sub theme to override the Language Switcher, to demonstrate how we can adjust the breakpoint at which the Language Switcher displays in full, per implementation, based on the other elements in the Global Header.

sass/cd/cd-header/_cd-global-header.scss
```diff
@@ -5,13 +5,49 @@
...
+    [dir="rtl"] & {
+      float: right;
+    }
+  }
+}
+
+// If there are multiple menus, separate them with pipes and margins */
+/* This should correspond with the dropdown menu breakpoint */
+.js {
+  .cd-global-header__actions > * + nav {
```
left23, line 34: The wildcard selector is because the Language Switcher element is a `div` (Drupal block) whereas the other elements in `.cd-global-header__actions` are `nav` (Drupal Menu blocks).

sass/cd/cd-header/_cd-language-switcher.scss
```diff
@@ -6,123 +6,203 @@
...
-      padding-left: 12px;
-    }
-
-    &::after {
-      position: absolute;
-      right: 0;
-      width: 14px;
-      height: 14px;
-      content: "";
-      background-image: url("../img/icons/selected--highlighted.svg");
```
left23, line 94: This is the checkmark icon that was beside the active language in the dropdown. The SVG can also be removed.

sass/cd/cd-header/_cd-language-switcher.scss
rupl: Thanks for highlighting this. It took me a while to find these styles and reset in all the places I needed. By coincidence, FR/ES/AR languages produced overflow at the default 768px breakpoint when I tested locally using the menus enabled on our CD Demo site (lang x6, help, user).

sass/cd/cd-header/_cd-user-menu.scss
```diff
@@ -16,13 +16,7 @@
...
   // @todo add proper classes to make it more robust.
   svg + span {
-    [dir="ltr"] & {
-      margin-left: 10px;
-    }
-
-    [dir="rtl"] & {
-      margin-right: 10px;
-    }
+    margin-inline-start: 10px;
```
rupl, line 19: love it!

sass/cd/cd-header/_cd-user-menu.scss
```diff
@@ -68,16 +62,16 @@
...
+  // &::after {
+  //   position: absolute;
+  //   top: 50%;
+  //   right: -1px;
+  //   width: 2px;
+  //   height: calc(var(--cd-global-header-height) / 3);
+  //   content: "";
+  //   transform: translateY(-50%);
+  //   background: var(--brand-grey);
+  // }
```
rupl, line 74: i'm ok to remove this completely as I made clear in the other PRs ;)

sass/cd/cd-header/_cd-user-menu.scss
```diff
@@ -215,7 +207,46 @@
...
+.cd-global-header .cd-global-header__user-menu:last-child .cd-global-header__dropdown {
+  right: 0;
+  left: initial;
+
+  [dir="rtl"] & {
+    right: initial;
+    left: 0;
+  }
+}
+
```
rupl, line 231: ok I see that I overlooked this when doing the flexbox conversion in #358 

I think we can accomplish this with the same logical properties since flex keeps it flush in my other PR.

sass/cd/cd-header/_cd-language-switcher.scss
```diff
@@ -6,123 +6,182 @@
...
-    margin: auto; // Center vertically in macOS 10
-  }
-
   &:focus {
     outline: 3px solid var(--brand-primary--light);
   }
 
   @include tablet {
-    padding: 0 0 0 20px;
+    padding-inline: 1.5rem;
```
rupl, line 25: Not to myself, should this be `padding-inline-start` to only apply to LTR left?

rupl: Nope, just a change in styles. This matches the default styles which also apply padding to both inline directions.

sass/cd/cd-header/_cd-user-menu.scss
rupl: Nevermind, this positions the dropdown and is a nice cosmetic to retain.

rupl:
Whew! Ok I think we're good to go. I will let this sit and review it with fresh eyes one more time tomorrow.

sass/cd/cd-header/_cd-global-header.scss
```diff
@@ -15,3 +15,20 @@
     float: left;
   }
 }
+
+// If there are multiple menus, separate them with pipes and margins */
+/* This should correspond with the dropdown menu breakpoint */
+.cd-global-header__actions > * + nav {
```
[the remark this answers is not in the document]
rupl: I got rid of all the selectors ;)

sass/cd/cd-header/_cd-language-switcher.scss
rupl: The block has been reduced quite a bit by eliminating pseudo-element pipes. I did leave the component in the CD Site PR but also copied it to the sub-theme starterkit (disabled by default) so that it's immediately available to override on any site.

Once we finally get rid of Sass and have each component defined as a Drupal library, we can change our commented-out examples in the theme info file to use Language Switcher so that overriding the breakpoint is a couple keystrokes.

PR #356, opened 2022-12-08
CD-254 User menu

left23:
Branches off of `CD-231` #350 

## Types of changes

- :heavy_check_mark: Improvement (non-breaking change which iterates on an existing feature)
- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- Security update (dependency updates, or to fix a vulnerability)

## Description
This updates the User Menu styles to accommodate recent changes to the Language Switcher display from #350
This PR also removes the account name as the User menu dropdown button and all the variations on the User menu previously (see [doc](https://docs.google.com/document/d/1nKZs8LMDwBKDRECdrEDdWp-47saRpCsUO7tLQsOXAXY/edit#heading=h.cr7oiiwqywk1)), simplifying things to a single display option where the Log out menu item is always in a dropdown regardless of the number of other menu items in the User account.

## Motivation and Context
Previously we had multiple variations on the User menu display depending on the other items in the user menu. Simplifying this means we have one implementation method that is used regardless of the different User menu items per property.

## Expected behavior
The Log in menu should display as it always has, as the last item in the Global Header.
Once logged in, the Log out menu item is available in a dropdown, with "My account" as the parent. Any additional User menu items will be placed in the dropdown as well, including the "My account" item that comes from the Drupal User module. This item is renamed to "Edit my account" using a preprocess function.

See [doc](https://docs.google.com/document/d/1nKZs8LMDwBKDRECdrEDdWp-47saRpCsUO7tLQsOXAXY/edit#heading=h.7y1ie8rvhmjy) for configuration instructions and screenshots

## Steps to reproduce the problem or Steps to test

  1. Check out this PR and use the same configuration for `common-design-site` when testing the Language Switcher's new display https://github.com/UN-OCHA/common-design-site/pull/225
  2. Follow the instructions in the doc. If testing on a site that has the previous set up, rename the parent User menu item from `%username%` to "My account"

## Impact
Depending on how the User menu is configured, this might require configuration changes. Without it, we might see `%username%` as a Menu item title. This should be changed as needed to "My Account"

## PR Checklist

- [x] I have followed the Conventional Commits guidelines.
- [ ] I have run local tests and the tests pass.
- [x] I have linted my code locally.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have made changes to the sub theme to reflect those in the base theme

PR #360, opened 2022-12-20
[CD-254] User menu rebased

rupl:
This is achieving the goals of #356 — go there for all notes about this PR.

This PR exists to make sense of the chaos I caused by squashing and merging #350. I'll probably open a third PR to merge into dev once I figure out what needs to merge and what is duplicate.


rupl:
Whew, ok that diff looks much more sensible. Comparing to #356 shows me an almost-identical diff (I did intentionally change one or two things).

PR #366, opened 2023-01-11
build(deps): bump stylelint-config-standard to match Drupal 9.5.x

rupl:
## Types of changes

## Description
Avoid linting errors when committing CSS changes to individual implementations.

## Motivation and Context
After upgrading a Drupal site to 9.5.x I noticed that new CSS changes could no longer pass linting. This seems to fix it.

## Expected behavior
Change and commit CSS error free.

  1. Install this branch on a 9.5.x site — or — manually bump the subtheme of a 9.5.x site to have `stylelint-config-standard@29`
  1.
  1.
  
  
## Impact
This will prevent the theme from working on 9.4.x or older (while still allowing 8.8.x or higher) but since 9.4 is EOL I don't consider it a breaking change.

## PR Checklist

### DIFF
--- a/sass/cd/cd-header/_cd-user-menu.scss
+++ b/sass/cd/cd-header/_cd-user-menu.scss
@@ -51,10 +51,6 @@
   background: transparent;
   line-height: var(--cd-global-header-height);
 
-  @include tablet {
-    max-width: none;
-  }
-
   &:hover,
   &:focus {
     text-decoration: underline;
@@ -77,6 +73,9 @@
     }
   }
 
+  @include tablet {
+    max-width: none;
+  }
 }
 
 .cd-user-menu__dropdown {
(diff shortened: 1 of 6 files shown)


lazysoundsystem (approved):
I ran into this on https://humanitarian.atlassian.net/browse/OPS-8923 - I updated a *whole heap* of dependencies but not this one, and got stumped when the linting failed, after many other `npm run` commands also failed.

sass/cd/cd-header/_cd-user-menu.scss
```diff
@@ -141,17 +140,6 @@
   }
 }
 
-/* Last block in Global Header should have its dropdown right aligned */
```
left23, line 144: @rupl I got to this too late, but I'm wondering if removing this CSS rule was intentional?

rupl: @left23 yes intentional. The entire block of styles was duplicated and the linter told me about the duplicated selector when I was finalizing the PR. I can't comment directly on the line because it's not close enough to the diff but here's the block:

https://github.com/UN-OCHA/common_design/pull/366/files#diff-e3e7a9f9f60276a5348104bf00f644b49dadadc4a3c302195dc1f6135e40098cR132

I just double-checked and it does seem like it was two identical blocks of styles. So I think we're good.

PR #373, opened 2023-01-26
CD-392: remove Sass

rupl:
## Types of changes

- :heavy_check_mark: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Description
Totally removes Sass from both base/sub-themes. All code which previously relied on Sass has been migrated to Drupal Libraries.

## Motivation and Context
This significantly reduces overhead department-wide:

1. Maintainers: We have fewer dependencies and thus fewer security updates.
2. OPS/deployments: Up until now we required a "builder bot" for all Drupal websites, because the possibility existed that they ran Sass in order to compile production-ready assets. Removing the builder bot will mean that all Drupal deployments are simpler and faster. 
3. Theming: People building sites have fewer tools to learn. Non-Drupal sites benefit from a more modular CD.

  1. Install the base-theme on a site.
  2. Override the necessary templates.
  3. Observe whether anything got knocked out of place.
  
## Impact
The following list should all be considered BREAKING CHANGES. **If the impact of one particular change ends up being too large and affects all sites, perhaps we need to remove the disruptive modifications.**

- CSS is now driven entirely by Drupal Libraries. Overrides, extends, and replacements are all available by targeting the new libraries in the base-theme: `cd-resets`, `cd-header`, `cd-footer`, `cd-other`
- The CSS groups have been refactored, so that CSS it output in a different order. For example the CD resets are using `base` while most components use `component` — previously our compiled Sass all lived inside `theme` which is the last group to be output in source order.
- All packages used within the Sass task were removed: `autoprefixer`, `postcss`
- All SCSS-related linting dependencies/config were removed.
- The npm commands related to linting have been renamed: `css-lint` and `css-lint-fix`
- The default font Roboto is defined in HTML instead of using `@import` within the compiled Sass bundle.
- Markup for social menu in Footer was refactored to use Flexbox
- Physical dimensions were refactored to use [logical properties and/or values](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties), i.e. `border-left` for LTR turned into `border-inline-start` — meaning the associated RTL rules could be eliminated. This refactor wasn't absolute; many instances still remain but since this has an impact on IE11
- People still using Sass in projects whose deployments are managed by IMB will need to check their compiled Sass into version control to avoid the need for the builder bots. Composer files will need to be updated to remove any references to Sass commands (or the old linting commands).

## PR Checklist


css/brand.css
```diff
@@ -41,6 +41,18 @@
   --cd-container-padding-tablet: 1.5rem; /* 24px */
   --cd-container-padding-xlarge: 2.5rem; /* 40px */
 
+  /**
+   * Breakpoints
+   *
+   * Mostly to document the old Sass vars, these values aren't used directly in
+   * CSS when needed for a media query, because custom props don't work in MQs.
```
rupl, line 48: Whoops, I accidentally commented on the commit instead of the PR. Reiterating here: I think the original wording is correct: the _values_ are used directly in media queries, e.g. `768px` instead of `var(--cd-bp--md)`

left23: Ah ok, I'll revert

left23 (approved):
Looks great.
I did a visual review and checked out the code. I saw issues you already sorted here https://github.com/UN-OCHA/common-design-site/pull/251 and attempted to run VRTs but the failed to produce the HTML report (OPS issues). Approving this so as not to block.

rupl:
The only two VRs I know are:

- social footer spacing
- today I adjusted the Search form to use flexbox, and it caused a change in how the focus on the search input displays. Search button has to be clicked to see it.

templates/navigation/menu--main.html.twig
```diff
@@ -28,22 +28,28 @@
...
+       classnames begin with 1 instead of 0. #}
+    {% set menu_level_cosmetic = menu_level + 1 %}
+    {% set component = 'cd-nav-level-' ~ menu_level_cosmetic %}
 
     {%
       set menu_classes = [
         'menu',
-        menu_level > 0 ? 'cd-main-menu__dropdown',
+        'cd-nav__menu--level-' ~ menu_level_cosmetic,
+        menu_level > 0 ? component ~ '__dropdown',
```
left23, line 40: @rupl I set up a branch to deploy to Feature site for VRT, and discover the steps where sass is still attempted to be compiled. It also served to run the e2e tests (which I couldn't get working locally) See https://github.com/UN-OCHA/common-design-site/pull/250
I think the highlighted code above breaks the tests `MainMenuDropdown` as it changes the class the test looks for. This means any sub themes with tests copied from the base theme will have to be updated too.

rupl: Ah yeah, right... I had remembered a couple tests but the main nav ended up taking several tries and I let the testing slip. Thanks for catching it! Thanks for setting up a better v8 PR too.

rupl: Fixed by f85cd50

PR #383, opened 2023-03-23
[CD-454] :focus-visible styles

rupl:
# CD-454, CD-357

## Types of changes

## Description
Adopt `:focus-visible` styles in Header/Footer

## Motivation and Context
The focus styles are less obtrusive for mouse users, which will hopefully mean that teams stop overriding them with harmful resets that remove the styles entirely, causing accessibility issues with the modified websites.

Additionally, keyboard navigation is more attractive and hopefully that translates to a UX improvement.

The Search dropdown also uses the same `:focus-visible` styles so that it shows focus during keyboard nav. This issue was previously ticketed in CD-357.

## Expected behavior
Mouse/touch users will no longer trigger focus styles when toggling dropdowns. The focus styles have been consolidated and are using CSS Vars so that they can be restyled in a unified fashion (for example when picking branding colors at the beginning of a project, or if we ever offered a high-contrast mode).

  1. Interact with each UI widget in the Global Header via mouse
  1. Interact with each UI widget in the Global Header via keyboard nav  

See ticket for videos of before/after.

## Impact
There are no steps to upgrade unless someone has modified the CD itself within their project-specific files.

## PR Checklist

### DIFF
--- a/components/cd/cd-footer/cd-footer-copyright.css
+++ b/components/cd/cd-footer/cd-footer-copyright.css
@@ -23,10 +23,6 @@
   }
 }
 
-.cd-footer-copyright__text a:focus {
-  outline-offset: 1px;
-}
-
 .cd-footer-copyright svg {
   flex: 0 0 38px;
   width: 38px;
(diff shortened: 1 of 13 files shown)


rupl:
ooh I forgot to mention: the OCHA Services logo has been nudged 3 pixels inward to accommodate the outline. I did that on purpose.

left23 (approved):
Looks great, approving.

PR #391, opened 2023-04-28
[CD-462] modernize layouts

rupl:
# CD-462

## Types of changes

## Description
Removes use of CSS Grid for the page layout and instead uses Flexbox. The code for the page is contained within its own file and uses a unique classname since it serves a unique purpose (mainly to create a "sticky footer").

Modernizes the sidebar layouts while providing a "universal layout" for older browsers. The universal layout is unconditional and only contains CSS 2.1 properties and values, with a few optional media queries to space things out.

The new modern layout is wrapped in `@supports` and requires three feature detections to pass:

- CSS Flexbox
- CSS Logical Properties and Values, e.g. `padding-inline-start: 1rem`
- CSS Custom Properties, e.g. `color: var(--brand-primary)`

Provided all three are supported by a given browser, it will render the modern layout with significantly less code, and older browsers no longer require any polyfills or JavaScript to be able to render a basic layout. It does this will relatively little code and also provides robust RTL out of the box.

The classnames were overhauled to more closely follow BEM. The "faux column" layout has been replaced by the modern layout by way of replicating the BEM classnames.

I have attempted to apply this to the CD Demo page as well. @left23 a question for you: are we actively using the files ending in `.html` or are they supposed to be vanilla examples for other frameworks? If they are just examples, would you be opposed to removing them and simply relying on Twig, so that we have a single source of truth? If they are being used in the code, I can't see how and will appreciate your advice on how they're used.

  1. Check branch out, visit all the URLs in the main nav
  1. It might help to expand the contents of the sidebar blocks. I added more paragraphs and an unordered list so that I could test the normal/compact modes
  
  
## Impact

If someone overrode any of the following templates, they'll need to reflect the new classnames adopted by our CD base-theme templates:

* `node--full.html.twig`
* `maintenance-page.html.twig`
* `page--404.html.twig`
* `page--demo.html.twig`
* `page--facets.html.twig`
* `page.html.twig`

## PR Checklist

# CD-460

GitHub Action will only run during PRs once this branch gets merged. No change to the Action itself.


left23 (approved):
Looks good @rupl Approving. I added a few notes to the JIRA issue with screenshots. I'm happy to review again if needed.

templates/layout/page--404.html.twig
```diff
@@ -6,9 +6,9 @@
 
       {{ page.page_title }}
 
-      <div class="cd-layout-content-wrapper">
+      <div class="cd-layout">
 
-        <div class="cd-layout-content">
+        <div class="cd-layout__content">
           {{ page.content }}
         </div>{# /.layout-content #}
```
left23, line 13:
```suggestion
        </div>{# /.cd-layout__content #}
```

PR #397, opened 2023-05-31
[CD-476] Improve screen reading

rupl:
## Types of changes

## Description

Improves screen reader performance by eliminating redundancies, or consolidating relevant properties. The results mean the screen reader announces far fewer words, while still announcing all the important roles and labels.

### Global Header menus

Each navigation section used to have a visually-hidden heading referenced with `aria-labelledby`. That markup caused every heading to effectively be announced twice (once for `aria-labelledby` and once for the heading itself). All of the headings have been removed, and the `aria-label` attribute is used in place of `aria-labelledby`, with Twig-translated strings as each label. The labels were chosen to be very short, with the assumption that the word "navigation" gets read afterwards:

| Entering | Leaving |
|:--|:--|
| "Language, navigation" | "end of, Language, navigation" |
| "Help, navigation" | "end of, Help, navigation" |
| "Account, navigation" | "end of, Account, navigation" |

### Help/Account menus

The other two menus in the Global Header are identical in structure to each other, and both always render as a dropdown whose button used to be the only top-level item of a two-level nested list.

Their templates were modified to eliminate the top-level `<ul><li>` surrounding the toggle buttons. If the menu has no descendant items, the `<ul>` will not be rendered at all. With the list and list-items removed around the toggle button, the screen reader thus announces two fewer items before offering the user the chance to open the dropdown, and a total of three announcements were removed if the person wishes to pass by the menu completely.

- Help, navigation
- ✂️ ~List, 1 item~
- ✂️ ~List item 1 of 1~
- Help menu pop up button
- _[if you open the menu, it will begin a new list and go through each item]_
- ✂️ ~end of list~
- end of, Help, navigation

### Site logo

It now announces the same in mobile/desktop by marking the image tag as hidden. In its place we have `visually-hidden` text of the site name.

### Landmark roles

Header/Main/Footer are all marked up identically now, with `aria-label` piped through Twig translation. Sidebars retain their `complementary` roles (no `aria-label` present).

### Footer

- Footer/Social nav are now `<nav>` with explicit `aria-label` translated by Twig. visually-hidden headings have been removed.
- OCHA Mandate avoids announcing "Sketch. empty artbox" and the SVG will now read as the full-text of UN-OCHA spelled out.
- Copyright notice has a header to announce it now.

  1. Use the OS, browser, and screen reader of your choice. For testing purposes, you might want to slow down the speech in your OS settings.
  1. Enable the screen reader and give the page a refresh to start from the beginning.
  1. Compare readout to notes above, and to [production](https://web.brand.unocha.org/)
  
## Impact
No breaking changes expected. However, without some manual modification, existing sites may not receive all the benefits  contained in this PR:

- If any of the block/links/menu templates for the relevant menus were overridden, the changes in this PR should be incorporated. Detailed procedures or advice can be given on a case-by-case basis.
- If `block--system-branding-block.html.twig` was overridden, its changes need to be incorporated.
- If any of the `page.html.twig` variants exist, they will need some attributes added to match the base-theme. Same for maintenance page.

## PR Checklist

### DIFF
--- a/components/cd/cd-header/cd-user-menu.css
+++ b/components/cd/cd-header/cd-user-menu.css
@@ -10,6 +10,7 @@
   margin: 0;
   padding-inline-start: 0;
   list-style: none;
+  position: relative;
 }
 
 .cd-user-menu ul {
--- a/components/cd/cd-header/cd-user-menu.css
+++ b/components/cd/cd-header/cd-user-menu.css
@@ -1,5 +1,9 @@
 /**
- * Common Design: User Menu
+ * Common Design: User Menus
+ *
+ * User menus are the dropdowns in the Global Header, such as Help or Account.
+ * Site configuration can affect how many of them are displayed, but they all
+ * get styled in a similar fashion.
  */
 
 .cd-user-menu {
@@ -21,11 +25,12 @@
  *
  * @TODO: add proper classes to make it more robust.
  */
+.cd-global-header__user-menu svg + span,
 .cd-user-menu svg + span {
   margin-inline-start: 10px;
 }
 
-/* Other icons, for example user or help icon. */
+/* Other icons, for example Help/Account icons. */
 .cd-user-menu .cd-icon:not(.cd-icon--arrow-down) {
   width: 14px;
   height: 14px;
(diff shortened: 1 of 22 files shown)


rupl:
@lazysoundsystem is it possible you could give this a reading? I'm not expecting super detailed feedback, but as long as you don't hear anything too strange I am considering it good enough to ship.

lazysoundsystem (approved):
I got the screen reader to go through once, but couldn't manage to make it repeat and missed a few things as it was faster than I could keep up with... I tried with Orca, which I've had better luck with before, and the firefox Pericles extension.

But I didn't notice anything jarring and the changes all look reasonable, so approving.

PR #403, opened 2023-06-15
fix: prevent CD Form from applying outside <form> elements

rupl:
# CD-473

## Types of changes

## Description
We had some collisions with [Admin Toolbar](https://www.drupal.org/project/admin_toolbar) and decided the easiest thing to do is prefix our selectors with `form`. All of our use-cases for this component involve true `<form>`s so it seems to be a safe modification.

  1. Test on a Drupal install that uses forms (CD Demo, Sesame, RW)
  1.
  1.
  
  
## Impact
We increased specificity so any CSS overrides might need to be adjusted. Simply add `form` to the selector in the appropriate location and the override will again take effect.

## PR Checklist

### DIFF
--- a/components/cd-form/README.md
+++ b/components/cd-form/README.md
@@ -1,26 +1,19 @@
 # Form
 
 ## Purpose and Usage
-CSS-only component for basic form element styles.
-The current approach is defaulting to browser styles with some CD styles added over
-time as we work on consistency with other properties (RW, HID).
-This means we don't expect it to look the same across every browser.
+CSS-only component for basic form element styles. The current approach is defaulting to browser styles with some CD styles added over time as we work on consistency with other properties (RW, HID). This means we don't expect it to look the same across every browser.
 
 Also endeavour to keep this devoid of rules that have implication to layout,
-where possible.
-
-This can expand as needed. Currently, there are generic and Drupal-specific
-selectors.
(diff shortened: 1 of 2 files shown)


rupl:
@left23 NOT urgent, but I wanted to see if you know of any overrides to this component in Sesame or RW that could be affected. I believe those were the primary projects that resulted in this component. If there are overrides, I'd put together fixes that incorporate these changes. All you have to do is highlight potential issues.

I'll be happy to re-ping about this in a month or so; that was when UNO is wrapping up, IIRC?

left23:
> @left23 NOT urgent, but I wanted to see if you know of any overrides to this component in Sesame or RW that could be affected. I believe those were the primary projects that resulted in this component. If there are overrides, I'd put together fixes that incorporate these changes. All you have to do is highlight potential issues.
> 
> I'll be happy to re-ping about this in a month or so; that was when UNO is wrapping up, IIRC?

@rupl Good question and I am not sure so I would need to look into it. I can do so quickly enough though to at least confirm whether it might be an issue and then you'll know how to proceed. Leave it with me for now please. I'll look into it ASAP.

rupl:
You really don't have to prioritize this! It can wait until UNO is wrapped. I just released a new CD today without the PR and Ralf's problem is solved inside his sub-theme, so there's no rush.

PR #405, opened 2023-06-29
[CD-482] CD Social Links - label and a11y

rupl:
# CD-482

CTC = copy to clipboard.

## Types of changes
- :heavy_check_mark: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Description
- **Added a label to the Social Links:** we never had one before and UNO wanted one.
- **Refactored the CTC button:** the previous implementation was not accessible at all. It no longer announces the user feedback before the button, instead it uses `div[role=status]` to announce the feedback after clicking. The link has been converted to a button, and the share URL has been moved from `href` to `data-url`.

  1. Observe the label on CD Demo
  1. Turn SR on, when arriving at the CTC button it should ONLY say "Copy URL to clipboard" (translatable)
  1. When clicking CTC button, it should ONLY announce "URL copied to clipboard" (translatable)
  
## Impact
⚠️ **BREAKING CHANGE:** If you've implemented the CD Social Links and used the CTC button, you MUST update your markup to continue using the button.

## PR Checklist


rupl:
@left23 no review required, only FYI.

PR #411, opened 2023-07-03
[CD-481] card image gaps, ARIA roles, demo tweaks

rupl:
## Types of changes

## Description
There used to be a little gap under every Card image. By setting the image to `display:block` we eliminate that gap. Now the visual appearance will match whatever you wrote in code. That is, if you have a flex `gap:1rem`, and a Card container `padding-top:1rem` then it will truly be 2rem instead of a little extra.

  1. Check the demo

Before / after, with Flex item borders overlaid.

CD-481-1-before CD-481-2-after

## Impact
If you relied on an image to be `display:initial` (inline content) then you might have to enforce that explicitly now. Most cards should be unaffected since they use Flex by default.

```css
.cd-card__image img {
  display: initial;
}
```

## PR Checklist

### DIFF
--- a/components/cd-article/cd-article.html
+++ b/components/cd-article/cd-article.html
@@ -41,7 +41,7 @@
 
         </div>
 
-        <aside class="cd-layout__sidebar cd-layout__sidebar--first" role="complementary">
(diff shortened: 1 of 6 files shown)


rupl:
@left23 FYI

PR #427, opened 2023-09-14
[CD-489] CD Disclosure component

rupl:
# CD-489

## Types of changes

- :heavy_check_mark: New feature (non-breaking change which adds functionality)

## Description
Dedicated component for content disclosure (show/hide expanded content) with a very down-to-earth name.

  1. I've updated the CD Demo page with the new component. Feel free to try it there on your local. There are a couple commented-out attributes in the template to test different configurations (see README for full list).

## Impact
No impact because it's a new component.

## PR Checklist


common_design.libraries.yml
```diff
@@ -239,6 +239,15 @@ cd-date:
   dependencies:
     - common_design/css-vars-ponyfill
 
+cd-disclosure:
+  css:
+    component:
+      components/cd-disclosure/cd-disclosure.css: {}
+    state:
+      components/cd-disclosure/cd-disclosure-states.css: {}
```
rupl, line 247: I guess another high-level question: worth it to have this "states" file or combine into one component file? Drupal aggregates differently based on this libraries config, so I wanted to give it a spin. But if it seems unweildy, I'm happy to recombine.

components/cd-disclosure/cd-disclosure.js
```diff
@@ -0,0 +1,81 @@
...
+    },
+
+    /**
+     * Set up all Disclosure widgets found in the DOM.
+     *
+     * For our purposes, the `disclosure` element is the wrapped content itself
+     * and the button to toggle it is generated automatically by this script.
+     */
+    initializeDisclosures: function () {
+      let disclosures = document.querySelectorAll('.cd-disclosure');
```
orakili, line 22: We need to set an attribute or class like `cd-closure--processed` and change the selector to exclude processed elements otherwise they may be processed again when the script is loaded again via ajax for example.

components/cd-disclosure/cd-disclosure.js
```diff
@@ -0,0 +1,81 @@
...
+        if (!disclosure.getAttribute('id')) {
+          disclosure.setAttribute('id', disclosureId);
+        }
+
+        // Determine where to place the button.
+        let positionAfter = disclosure.dataset.cdDisclosureInsert === 'after' || false;
+
+        // Determine whether content is shown by default. We consider this the
+        // edge-case so omitting the attribute assumes it is hidden.
+        let contentVisible = disclosure.dataset.cdDisclosureShowByDefault === 'true' || false;
```
orakili, line 39: We should set an initial `cd-disclosure--is-collapsed` or `cd-disclosure--is-expanded` on the element based on this value otherwise those classes are only set when pressing the button.

Also we probably want to remove the attribute.

See also comment about classes above.

components/cd-disclosure/cd-disclosure.css
```diff
@@ -0,0 +1,12 @@
...
+ * CD Disclosure
+ *
+ * This component allows you to implement a show/hide button for content while
+ * retaining a portion of it that is always visible. Typical use-cases are to
+ * show a "teaser" but allow the user to expand a longer-form piece of text.
+ *
+ * @see https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/
+ */
+.cd-disclosure {
+  display: none;
```
orakili, line 11: Not sure about this simple default because that means we need to set the visibility explicitly (the rules with `display:block` in cd-disclosure-states) which may be problematic for the styling of the content because it means an implementor would have to override the rule if they want `display: flex;` for example instead of the `display: block`.

Instead I think we can have the expected behavior with only 1 rule and 1 class (assuming the removal of the `data-cd-disclosure-show-by-default` and the addition of the class as mentioned in the other comment below):

```css
.cd-disclosure:not([data-cd-disclosure-show-by-default="true"], .cd-disclosure--is-expanded) {
  display: none;
}
```

Related to that is hiding the content the best default when javascript is disabled? If not, we could a `.js ` before the rule(s).

components/cd-disclosure/README.md
```diff
@@ -0,0 +1,14 @@
...
+**Minimum requirements**
+
+- `.cd-disclosure` classname. JS will consider any element with this classname to be collapsible content.
+
+**Optional attributes**
+
+- `id`: Set your own custom HTML ID in Twig if you wish. The JS will find it and automatically configure the button to control that ID. If you omit the ID, JS will generate a random one in the form `cd-disclosure--XXXXXXXX`
+- `[data-cd-disclosure-show-by-default]`: set this attribute to `true` if you want the content to show by default, but still create a button that can toggle its visibility. If the attribute is omitted or set to anything except `true` then JS will default to `false`
+- `[data-cd-disclosure-label]`: provide a label for the button. You can combine this with Twig's translation filter to generate multilingual buttons based on page context and available UI String Translations inside Drupal.
+- `[data-cd-disclosure-insert]`: set this attribute to `after` to have the button inserted after your content. If you omit this attribute or supply any value besides `after` it will default to "before".
```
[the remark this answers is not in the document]
orakili: The 2 use cases we have right now are RW and UNO with the button after the content for a "show less/more" type of disclosure. That said I don't have a strong opinion on it but with the current behavior of having the content hidden by default, I would go with `before` as default.

common_design.libraries.yml
orakili: Looking at the number of rules (especially, see other comment about them), I don't think it's worthy to have 2 css files.

components/cd-disclosure/cd-disclosure.js
```diff
@@ -0,0 +1,81 @@
...
+
+        // Extract button label
+        let buttonText = disclosure.dataset.cdDisclosureLabel || 'Toggle content';
+
+        // Create the toggler button.
+        let disclosureButton = document.createElement('button');
+        disclosureButton.setAttribute('aria-controls', disclosureId);
+        disclosureButton.setAttribute('aria-expanded', contentVisible);
+        disclosureButton.innerText = buttonText;
+        disclosureButton.classList.add('cd-button');
```
orakili, line 49: This is missing some classes to be able to style the button (for example to add an arrow icon via `::after` for example) something like `cd-disclosure-toggler`?

components/cd-disclosure/cd-disclosure.js
```diff
@@ -0,0 +1,81 @@
...
+        // edge-case so omitting the attribute assumes it is hidden.
+        let contentVisible = disclosure.dataset.cdDisclosureShowByDefault === 'true' || false;
+
+        // Extract button label
+        let buttonText = disclosure.dataset.cdDisclosureLabel || 'Toggle content';
+
+        // Create the toggler button.
+        let disclosureButton = document.createElement('button');
+        disclosureButton.setAttribute('aria-controls', disclosureId);
+        disclosureButton.setAttribute('aria-expanded', contentVisible);
```
orakili, line 47: I think the button could benefit from an `ID` derived from the `disclosureId` so it can easily be targeted by other javascript (thought that can be done with a `button[aria-controls="ID"]`). (Use cases: RW and UNO which have extra logic for the scrolling to the top of the content).

orakili (changes requested):
Thanks. It's a good start.

components/cd-disclosure/cd-disclosure.js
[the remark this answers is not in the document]
rupl: I actually started out with data attribute only, but switched because one thing that all components in the CD have is that they have a top-level CSS classname, including CD Dropdown. That's why I landed on the solution as written. But I do agree that in terms of how JS functions, using the attribute makes sense.

I've added the data-attr back in, and keep the optional class in the example. The JS will look for the attribute, not the class.

If we ever offered some styling variant, or for example wanted to offer declarative transitions like slide-up/down, then I think we can agree that modifier classes are the established standard, both in our CD codebase and as a general industry convention.

components/cd-disclosure/cd-disclosure.js
```diff
@@ -0,0 +1,81 @@
...
+
+        // Determine where to place the button.
+        let positionAfter = disclosure.dataset.cdDisclosureInsert === 'after' || false;
+
+        // Determine whether content is shown by default. We consider this the
+        // edge-case so omitting the attribute assumes it is hidden.
+        let contentVisible = disclosure.dataset.cdDisclosureShowByDefault === 'true' || false;
+
+        // Extract button label
+        let buttonText = disclosure.dataset.cdDisclosureLabel || 'Toggle content';
```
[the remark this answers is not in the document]
rupl: I can't remember where I read it now, but I found guidance that specifically advises against changing button labels according to state. The example I saw involved a play/pause button for a video player, where the label is "pause" when video plays, but "play" when it's paused.

This isn't the page, but it also avoids changing labels and instead offers a visual cue based on `aria-pressed`: https://www.w3.org/WAI/ARIA/apg/patterns/button/examples/button/

Additionally, the WAI Disclosure pattern also uses a visual cue instead of textual, in the form of a chevron. We could use our `aria-expanded` attribute to provide such a visual cue:
https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-image-description/

At any rate, that's why I intentionally excluded the second label. But for now, to keep consistency with Dropdown I will add it in.

components/cd-disclosure/cd-disclosure.css
rupl: Really good point about the `.js` class. I'll take the above suggestions and consolidate into one file.

components/cd-disclosure/cd-disclosure.js
rupl: I'd assumed the button selector you mentioned would be how these things get targeted, but we can add another ID when we create them no problem 👍🏼

components/cd-disclosure/cd-disclosure.js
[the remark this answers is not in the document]
orakili: **Changing labels?**

A recommendation about not changing the label is mentioned in https://inclusive-components.design/toggle-button/ ("Changing labels" section).

The reasoning is sound, yet that feels strange for the use on RW/UNO use case with the button after the content to reveal long content.

I couldn't find really a pattern for a show more/less type of disclosure with the button after the content.

We should probably rethink how it's handled on those sites. Not sure if we want to tackle that as part of the work on this component because we then need to involve the product owners if we change the design/behavior.

As proposed, one option is to add a parameter for a second label like in the cd-dropdown so that this new component can be used almost as is instead of the cd-dropdown on those sites.

However while this parameter would be optional, it's a potential open door to bad practice if changing label is considered as such.

**Alternative**

Alternatively, we could keep the single label attribute and add some extra javascript on those sites to listen to changes on the `aria-expanded` attribute for example to switch the labels to keep the current behavior. They already have a bit of complementary logic that could be extended for that.

This way this `cd-disclosure` component could follow the "best" practices and those sites could keep their current behavior until revisited.

@rupl What do you think?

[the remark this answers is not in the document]
rupl: Nice find, yup that was probably the link I had in mind. I've read that page a couple times over the years.

Yes, we can go with the alternative you proposed. Out of curiosity, what do you have in mind to listen for the attribute's changed value? `MutationObserver`?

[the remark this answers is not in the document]
orakili: Yes, like https://github.com/UN-OCHA/unocha-site/blob/develop/html/themes/custom/common_design_subtheme/templates/modules/unocha_donors/unocha-donors-list--oct-top-donors.html.twig.

components/cd-disclosure/cd-disclosure.js
rupl: I was planning on re-using the attribute inside Dropdown `data-cd-processed` for consistency.

components/cd-disclosure/cd-disclosure.js
rupl: Added a new attribute and docs to README. It still outputs `cd-button` by default, but now you can zero it out with an empty attribute, or add space-separated classes to your heart's desire.

components/cd-disclosure/cd-disclosure.js
rupl: I'm using `cd-for--ID` where it's the full ID of the other element. I like this as a convention we could use in the future, where you can always "find" the toggler ID and vice versa. I took inspiration from HTML `for` attributes on form elements `<label for="my-html-id">Label</label>`

components/cd-disclosure/cd-disclosure.js
```diff
@@ -0,0 +1,81 @@
...
+     * For our purposes, the `disclosure` element is the wrapped content itself
+     * and the button to toggle it is generated automatically by this script.
+     */
+    initializeDisclosures: function () {
+      let disclosures = document.querySelectorAll('.cd-disclosure');
+
+      // All behavior is controlled by data attributes.
+      disclosures.forEach((disclosure) => {
+        // Set a random ID if none is found.
+        let disclosureId = disclosure.getAttribute('id') || this.idPrefix + Math.floor(Math.random(1) * 100000000);
```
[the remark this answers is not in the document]
rupl: tbh I like how linear the component's code is without extra methods, so I went with what I thought were sufficiently unique IDs. I've bumped it from 10 million to 1 trillion and kept the strings shorter by converting to hex. Is that ok?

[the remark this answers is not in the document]
orakili: Sure.

components/cd-disclosure/cd-disclosure.js
orakili: @rupl Bumping this comment.

To work well with the css rule in [components/cd-disclosure/cd-disclosure.css](https://github.com/UN-OCHA/common_design/pull/427/files#diff-f38e08f580263890eccc2b8e4683a8ac7107376e694ddd21f25a86b891bd1f1a), we should:

1. Remove the `data-cd-disclosure-show-by-default` if it is set to true
2. Set `cd-disclosure--is-collapsed` or `cd-disclosure--is-expanded` class depending on the value of the attribute

rupl: Got it, sorry I missed this. Should be good now.

orakili (commented):
@rupl I tested the component as a replacement of the `cd-dropdown` for the disaster description on RW and there are some inconveniences.

The main one is that to detect the expanded/collapsed state, I had to add a mutation observer on the class attribute but because there are 2 changes (ex: removal of the `is-collapsed` and addition of the `is-expanded`), the callback is called twice.

When using the `cd-dropdown`, I could observe the `data-cd-hidden` attribute. Changing its value only triggered 1 event and it was easy to compare the new and old values.

So, if possible, I think it would be nice to add such an attribute (could be something like `data-cd-disclosure-state="expanded/collapsed"`).

We could probably get rid of the `data-cd-disclosure-show-by-default` and instead indicate to set an initial `data-cd-disclosure-state="expanded"` to have he same behavior.

We could also reduce the code by removing completely the `is-collapsed/is-expanded` classes if you are ok with styling based on an attribute instead of classes.

So all in one, I think such an attribute could simplify the js and css code and make it easier to detect the changes on the disclosure element.

What do you think?

rupl:
> the callback is called twice

Ok yeah, that's annoying. TBH I only added classes for both states to get around the fact that it can be expanded by default. Perhaps we should only leverage one of them, `is-collapsed` since that's the JS enhancement (no-JS means it shows by default).

I like the shorter, more clear data attribute you suggested: `data-cd-disclosure-state="expanded"`, but it does sound like it's duplicating the `aria-expanded` attribute of the button (albeit on the expandable element itself). And ultimately our codebase convention is to style components with BEM; that means modifier classes for state.

I'll try to find a combo where `is-expanded` is eliminated, we only use `is-collapsed` and the data attribute is renamed and kept in sync.

rupl:
Ok, I did mostly what I predicted, except the new attribute is `data-cd-disclosure-expanded` — code-wise it means we can use the same boolean as for the button's `aria-expanded` attribute, instead of turning a boolean into an arbitrary string.

The CSS is quite simple now too!

orakili (approved):
Thanks, with the latest modifications, I could reproduce the behavior we had with the `cd-dropdown` on RW with a reasonable amount of code changes.

rupl:
@orakili just an FYI I tried out the "Squash and Merge" option on this PR, so all the commits are flattened into one. Just in case you were using one of the commit hashes in your RW/UNO WIPs.

PR #431, opened 2023-10-04
CD-505 e2e tests

left23:
## Types of changes

## Description
Update Puppeteer and Jest and make sure the tests run. Remove tests from base theme, in favour of having them in the sub theme only.

  1. Copy the sub theme, or just the test-related files, into a working project like common-design-site
  2. Follow the README steps to run tests. 
  3. Depending on the project set up, some tests may fail - for example on the common-design-site, we have not replaced "This is Customizable" text in the OCHA Services dropdown so the `OCHAServicesDropdown` test fails. If there is no search, or no second level items in the main navigation, those test will also fail.

## Impact
We can integrate e2e testing where the Common Design is being used, as a QA step.

## PR Checklist


common_design_subtheme/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "common-design-subtheme",
-  "version": "9.0.1",
-  "description": "OCHA Common Design sub theme for Drupal 9+",
+  "name": "common-design",
+  "version": "9.2.0",
+  "description": "OCHA Common Design base theme for Drupal 9+",
```
rupl, line 4: Whoops! I've been forgetting to bump the version recently. Keep the new version, but the sub-theme should probably be different name/desc than the base-theme

common_design_subtheme/README.md
```diff
@@ -100,8 +100,36 @@ general details about base-theme and instructions for compilation. There should
...
+# If you want to run a limited number of tests, specify a string with the -t
+# argument. It will parse all of the describe() blocks and only run tests when
+# it matches the string you supply.
+
+# All tests that include 'OCHAServicesDropdown'.
+npm run e2e -- -t 'OCHAServicesDropdown'
+
+```
+
+### PHP CS
```
left23, line 127: This was not needed for UNOCHA when I was testing the e2e tests from this PR there.
So I think it's specific to common-design-site and shouldn't be included in the base/sub theme readme.
I've added a comment there.

PR #437, opened 2023-10-11
[CD-507] update JS linting config

rupl:
# CD-507

## Types of changes

## Description
`npm run js:lint` is more forgiving, in that it still returns exit status 0 when it cannot match any files. I also changed where it looks for files (used to be `js/` now it's both `js/` and `components/`).

 The base-theme JS command has also been updated to include `components/` and that whole directory did have to be linted. I adjusted some linting rules according to my personal preference. I'm flexible if it's not palatable to the group.

## Steps to test

  1. Before checking out the branch, `npm run js:lint` on the sub-theme. It should show the following:

```
$ npm run js:lint

> common-design-subtheme@9.2.1 js:lint
> eslint '{js,components}/*.js'


Oops! Something went wrong! :(

ESLint: 7.32.0

No files matching the pattern "js/*.js" were found.
Please check for typing mistakes in the pattern.
```

If you run `echo $?` it will print the number `2`

  2. After checking the branch out, `npm run js:lint` and the error should be gone, and `echo $?` will output `0`
  3. Add a JS file with badly formatted code e.g. `d = 'something';` anywhere inside `components/` and make sure the linter still catches the errors.

## Impact
No impact unless you were running `npm run js:lint` in CI, but I don't think we do that anywhere. We can now start doing it and sites with zero custom JS will still pass.

## PR Checklist


left23 (approved):
Works as described, thanks!

PR #441, opened 2023-10-23
CD-500 Focus styles for main navigation

left23:
## Types of changes

## Description
Increases the opacity of the active item underline for better visibility, uses consistent focus-visible styles so all items in header have the same style when tabbing with the keyboard.

  1. Check out this branch and compare with https://web.brand.unocha.org
  2. The Active trail and Active item should be preserved but the opacity increased for better visibility, also on mobile
  3. The Main Menu keyboard tabbing should have a blue outline like the other elements, but the click focus should remain as before (we are distinguishing between focus and focus-visible now)

## Impact
Update the base theme with the latest release when this PR is merged and included.
This will correct a recent regression so if no overrides have been made in the sub theme, no work should be needed. If there are overrides in place to correct the opacity, these can be removed in favour of the changes in the base theme.

## PR Checklist

### DIFF
--- a/components/cd/cd-header/cd-language-switcher.css
+++ b/components/cd/cd-header/cd-language-switcher.css
@@ -18,10 +18,6 @@
   line-height: var(--cd-global-header-height);
 }
 
-.cd-language-switcher__btn:focus {
-  outline: var(--cd-outline-size) solid var(--brand-primary--light);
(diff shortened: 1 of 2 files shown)


rupl (approved):
Lovely!

PR #451, opened 2024-01-03
[CD-483] SDC

rupl:
# CD-483

## Types of changes
- :heavy_check_mark: New feature (non-breaking change which adds functionality)

## Description
Paving the way for SDC, plus one component to start off. All files for Drupal Libraries are now inside the `libraries` subdirectory, and `components` (which is a required directory for SDC) now holds SDC only.

Note: the diff looks huge, but it's only due to renaming the `components` directory to be `libraries`.

Use this branch in conjunction with https://github.com/UN-OCHA/common-design-site/pull/369

  1. Check this branch out inside an existing site and `drush cr`
  2. See that the demo page still works as intended.

## Impact

If someone has directly referenced any file inside `components` then their theme will break. Care must be taken to remove these manual references and replace them with [Drupal Libraries overrides](https://www.drupal.org/docs/develop/theming-drupal/adding-assets-css-js-to-a-drupal-theme-via-librariesyml#s-libraries-override). If the theme was already using proper `libraries-override` entries, there is no impact to the existing site.

To begin using SDC in an existing sub-theme, it's advisable to rename the existing `components` subdir to be `libraries` and update all your custom entries in `common_design_subtheme.libraries.yml` to point at the new directory. Then you can start putting your SDCs into the `components` directory.

## PR Checklist


libraries/cd-byline/cd-byline.html.twig
```diff
@@ -7,7 +7,7 @@
   <div class="cd-byline__wrapper cd-max-width [ cd-bumper ]">
 
     <div class="cd-byline__image cd-byline__image--rounded">
-      <img src="/themes/contrib/common_design/components/cd-byline/mark-lowcock.jpg" alt="Mark Lowcock">
+      <img src="/themes/contrib/common_design/libraries/cd-byline/mark-lowcock.jpg" alt="Mark Lowcock">
```
rupl, line 10: heh, a great example of something that SDC would fix! Nice catch.

libraries/README.md
```diff
@@ -0,0 +1,140 @@
+### Component QA status
```
left23, line 1: This whole file is less relevant, if I understand correctly. I guess it will go away eventually. Is there reason to edit it down a lot now?

left23 (approved):

libraries/README.md
rupl: I don't feel the need to edit it down for this PR. It's a good historical reference and still relevant to the Libraries themselves.

PR #458, opened 2024-01-16
[RW-826][CD-530] Reduce CLS in Global Header

rupl:
# RW-826 / CD-530

## Types of changes

## Description
Improves the Cumulative Layout Shift (CLS), a Core Web Vital that is very important to Google and its search results.

  1. Load branch and stuff.
  2. Using the devtools of your choice, do the following:
     - Disable caching, or start with a private browsing window
     - Throttle the network to something slow (so that you can see the effect)
  3. Load the page and observe that the Help/User menus consistently render to the right of the Global Header (in LTR). They should NEVER appear on the left, even momentarily.

Example video:

https://github.com/UN-OCHA/common_design/assets/254753/40800568-1446-4bb8-8a72-bcd988e24899

## Impact
If the `cd-global-header.html.twig` file is overridden in the sub-theme, it will need to have the new line of HTML inserted, or the OCHA Services menu will never display.

@left23 if you could double-check this aspect that would be most helpful. I reviewed all the base-theme and sub-theme templates and believe I got the impact instructions correct. But your confirmation would be most helpful. If you agree, go ahead and tick the last checkbox in the checklist.

## PR Checklist

### DIFF
--- a/libraries/cd/cd-header/cd-global-header.css
+++ b/libraries/cd/cd-header/cd-global-header.css
@@ -17,7 +17,7 @@
   gap: 8px;
 }
 
-.cd-global-header__inner .cd-ocha {
+.cd-global-header__inner #cd-ocha-services-container {
   order: -1;
 }
 
--- a/libraries/cd/cd-header/cd-ocha-services.js
+++ b/libraries/cd/cd-header/cd-ocha-services.js
@@ -3,8 +3,8 @@
 
   Drupal.behaviors.cdOchaServices = {
     attach: function (context, settings) {
-      // Move the OCHA services section to the header.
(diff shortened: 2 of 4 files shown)


left23 (approved):
This looks good! I tested locally on RW and I think the steps are accurate.
Most sites will have a `cd-global-header.html.twig` override file. From what I can see, all include an extend from `@common_design/cd/cd-header/cd-global-header.html.twig` so no negative impact expected. https://github.com/search?q=org%3AUN-OCHA%20cd-global-header.html.twig&type=code

rupl:
That was my thinking as well with the `extend` sub-theme template. Great, then the instructions are accurate. Thanks for the 👀

PR #459, opened 2024-01-16
OPS-9885 Change directory for select_a11y css after SDC reshuffle

left23:
Refs: OPS-9885

## Types of changes

## Description
After implementing SDC, there was one library `select_a11y` that was still trying to get its CSS from `components` folder.

  1. Compare https://web.brand.unocha.org/facets (which is not picking up the styles) with https://feature.commondesign-unocha-org.ahconu.org/facets (which is an old deploy and still picking up styles from components directory due to older base theme version)
  2. With the changes in this PR, the stylesheet paths are correct and the result should look like the current Feature site, once the base theme version is updated.

## Impact
This is a bug fix so anywhere Facets are used with select_a11y widget should display correctly after updating the base theme with this fix in place.

## PR Checklist

### DIFF
--- a/common_design.info.yml
+++ b/common_design.info.yml
@@ -46,12 +46,12 @@ libraries-override:
   select_a11y/select_a11y_facets.widget:
     css:
       theme:
-        css/style.css: components/cd-select-a11y/cd-select-a11y.css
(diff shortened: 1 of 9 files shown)


rupl (approved):
I found a few more 🙈

PR #461, opened 2024-02-07
CD-538 Prevent last menu item spanning full width when items wrap

left23:
Set flex grow on last child to 0 to prevent full width when wrapping
See CD-538 for screenshots

## Types of changes

## Description
When there are many menu items, and the viewport approaches 1024px but before the entire menu collapses into the dropdown, the last menu item might wrap and currently it spans 100% of available width (the width of the menu container). This is an edge case and we can see in on unocha.org where there are enough menu items in the top level main menu to cause this. This does not prevent the menu items from wrapping, but prevents the last menu item from spanning 100% width and is thus more pleasant visually and more usable.

  1. If testing on the Common Design, I've copied and pasted additional menu items into the main menu using the dev tools/inspector to emulate this behaviour, then reduce the viewport to near 1024px. Or check unocha.org, also reducing the viewport to around 1024px.

## Impact
This is a base theme change, so should not need any extra work in the sub theme unless there are already rules in place that affect the main menu or its items.

## PR Checklist

### DIFF
--- a/libraries/cd/cd-header/cd-site-header.css
+++ b/libraries/cd/cd-header/cd-site-header.css
@@ -64,6 +64,11 @@ now that we layout with flex. */
   .cd-site-header__nav-holder nav > ul > .cd-nav__item {
     flex: 1 1 auto;
   }
+
+  .cd-site-header__nav-holder nav > ul > li:last-child,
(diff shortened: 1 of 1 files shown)


rupl (approved):
Works as advertised!

PR #463, opened 2024-02-12
[CD-537] Remove OCHA Services from Global Header

rupl:
# CD-537

## Types of changes

## Description

We don't want the OCHA Services menu in the Global Header anymore, instead placing it in the main Footer in between the existing nav and the mandate/copyright.

It would be great to test this by dropping this branch of the base-theme **into existing projects** and verifying that the contents of the OCHA Services menu is unchanged, despite the new location in the footer. This is my CD Site with nothing but the base-theme modified:

CD-537-footer-ocha-services

## Impact

The idea here is that there is no impact to existing sub-themes. If there is an impact, I'd like to continue hacking at this branch until we're all happy it's safe to deploy as a patch or minor release.

The branch includes changes to the existing sub-theme so that the stub template isn't necessary for new sites. As the comment says, in a future major release we can remove the stub from the `cd-header` directory.

## PR Checklist


left23:
@rupl Is this an opportunity to better name the elements that currently have `cd-ocha-dropdown` included in their class names? It would be more self-explanatory if the div class names didn't imply behaviour that doesn't exist anymore.

rupl:
I had considered it. The problem is that if those classes were used to customize anything then changing classnames will break existing sub-themes.

The safer option was to leave them alone, but I suppose we can peek through the few sites we don't maintain and see how disruptive it will be.

libraries/cd/cd-footer/cd-ocha.css
```diff
@@ -1,102 +1,105 @@
...
   margin: 0;
   padding: 0;
 }
 
+/**
+ * backwards-compat: the .cd-ocha-dropdown class is legacy, but is used by lots
+ * of sub-themes which customized their OCHA Services links. If these classnames
+ * get removed then the styling of the first column will look strange.
+ */
+.cd-ocha-services__link,
```
rupl, line 64: Classnames have been revamped to remove the reference to a dropdown, with the exception of these three rules accompanied by the comment.

libraries/cd/cd-footer/cd-ocha.css
```diff
@@ -0,0 +1,153 @@
+/**
+ * Common Design: OCHA Services
+ */
+.cd-ocha-services {
+  @media screen and (min-width: 768px) {
+    position: relative;
+    margin-block-end: 2rem;
```
left23, line 7: I think this should match the `.cd-footer__section` of 3rem so sections have a bit more space between them - see previous comment above.

left23 (changes requested):
Looks good!
I've left some comments to do with the vertical rhythm changing.

And the e2e tests need attention to reflect the removal of the OCHA Services dropdown:

 FAIL  _tests/OCHADropdown.test.js (21.835 s)
  OCHAServicesDropdown
    ✕ should expand when clicked (139 ms)
    ✕ should contain up to four links in the Related Platforms section (10001 ms)
    ✕ should NOT contain links with default text in the Related Platforms section (41 ms)
    ✕ should contain eight links in the Other OCHA Services section (10000 ms)
    ✕ should contain specific links in the Other OCHA Services section (224 ms)
    ✕ should include a button to UNOCHA.org DS list (38 ms)

libraries/cd/cd-footer/cd-footer.css
```diff
@@ -66,7 +66,7 @@
 .cd-footer__section {
   width: 100%;
   margin: 0 auto;
-  padding-bottom: 3rem;
+  padding-block-end: 2rem;
```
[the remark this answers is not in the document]
rupl: I was just trying to tighten things up a little bit. I shouldn't change layout without consensus. I will revert it.

rupl:
@left23 I believe it's good to go now. Removed one test related to the dropdown, renamed all the things, and updated selectors to match our CSS rules (where either `.cd-ocha-services__xxx` or `.cd-ocha-dropdown__xxx` exists, depending on age of the sub-theme)

left23 (approved):
Great. Renaming the OCHADropdown.test.js filename to remove the "dropdown" part might be timely too. 
Approving.

rupl:
Ah good catch on the filename. My OCD wants to force push the rename into the commit 👿 but the other PR already references the first hash. Oh well.

lazysoundsystem:
Not sure where best to ask this, so do move it if necessary.

Surprised to see the behavior of the OCHA services menu on ODSG when not logged in - when logged in, it's as it was before.
image
and
image

@rupl should I be investigating further?

rupl:
That's related to https://github.com/UN-OCHA/common_design/pull/458, not this PR, and I believe a cache-rebuild will jiggle it back into place. Either that, or I messed up the ODSG upgrade. However I just checked and its [Global Header template is indeed a clean copy](https://github.com/UN-OCHA/odsg8-site/blob/c85ef50966c7f087ec49fbdc0aa492bc09e25b09/html/themes/custom/common_design_subtheme/templates/cd/cd-header/cd-global-header.html.twig).

I'm happy to put the PRs together if necessary, but if you see this on another site, the upgrade instructions (adding one line of HTML) are here: https://github.com/UN-OCHA/common_design/releases/tag/v9.3.1

rupl:
@lazysoundsystem FWIW I loaded https://odsg.unocha.org/welcome?destination=/dashboard and had to refresh twice: first refresh it "broke" and second it got restored. After that, it stayed put in the top left.

lazysoundsystem:
Thanks @rupl - looks like it's all fixed now.

PR #472, opened 2024-03-12
fix: Add an element id to the outer status messages wrapper div, so backstop can target it.

cafuego:
In theory this should work, as there ought to not be another status messages wrapper.

Refs: OPS-9741

## Types of changes

-  Improvement (non-breaking change which iterates on an existing feature)

## Description

Add an ID to the main wrapper for status messages, so we can target it for VRT using backstop, so they do not mess with the screenshot diffs.

  1. Run VRT as auth user.
  2. See login-link-used messages in the screenshots.
  3. Oops.

## Impact
In theory this breaks nothing and fixes something.

## PR Checklist

### DIFF
--- a/templates/misc/status-messages.html.twig
+++ b/templates/misc/status-messages.html.twig
@@ -27,7 +27,7 @@
 
 {{ attach_library('common_design/cd-alert') }}
 
-<div id="messages" data-drupal-messages>
+<div data-drupal-messages class="messages-list">
 {% block messages %}
 {% for type, messages in message_list %}
   {%
--- a/templates/misc/status-messages.html.twig
+++ b/templates/misc/status-messages.html.twig
@@ -27,7 +27,7 @@
 
 {{ attach_library('common_design/cd-alert') }}
 
-<div data-drupal-messages>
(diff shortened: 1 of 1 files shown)


rupl (changes requested):
I actually have a PR open right now that would break this assumption. The OCHA AI Chat module can inject inline status messages. Is there a problem with targeting the pre-existing Drupal selector in VRT config? `[data-drupal-messages]`

rupl:
If that Drupal selector won't work, can we add a unique-enough classname instead of ID, which doesn't cause problems if it does manage to be output twice on rare occasions?

templates/misc/status-messages.html.twig
```diff
@@ -27,7 +27,7 @@
 
 {{ attach_library('common_design/cd-alert') }}
 
-<div data-drupal-messages>
+<div id="messages" data-drupal-messages>
```
cafuego, line 30:
```suggestion
<div data-drupal-messages class="messages-list">
```

rupl (approved):

PR #485, opened 2024-05-15
chore: fix unexpected error

lazysoundsystem:
Refs: CD-555

## Types of changes

## Description
Rewritten an if-clause

  1. It's only triggered on very few 404s, I tried visiting urls similar to those which gave this error but couldn't reproduce it.
  2. To test, check the syntax.

## Impact
How does the proposed change impact existing implementations? What action is needed to accommodate these changes?
No action needed.

## PR Checklist

### DIFF
--- a/common_design.theme
+++ b/common_design.theme
@@ -499,11 +499,13 @@ function common_design_preprocess_html(&$vars) {
   $status_code = \Drupal::requestStack()->getCurrentRequest()->attributes->get('exception');
 
   // Add body classes.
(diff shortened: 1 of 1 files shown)


rupl (approved):
Still works like usual. I stared at this quite a while trying to figure out what the functional difference is. I still don't see it, but if you find it to reduce errors, then all good by me 👍🏼

lazysoundsystem:
I'm also a bit unsure on this: https://www.php.net/manual/en/language.operators.precedence.php says that `==` has higher priority than `&&` but I still don't see how it could get as far as checking the statuscode of an exception. And most of the time, I don't think it does.

PR #493, opened 2024-11-28
chore: avoid lazy loading for print

lazysoundsystem:
Refs: OPS-11125

## Types of changes

## Description
Remove lazy loading before printing.

https://humanitarian.atlassian.net/browse/OPS-11125

## Impact
This adds a snippet of js to all pages, to avoid printing lazy-loading images. This may be too much.

## PR Checklist

- [ x ] I have followed the Conventional Commits guidelines.

### DIFF
--- a/js/cd-print.js
+++ b/js/cd-print.js
@@ -1,0 +1,19 @@
+/**
+ * @file
+ * Print helper.
+ */
+
+((Drupal) => {
+  'use strict';
+
+  Drupal.behaviors.cdPrint = {
+
+    // Remove lazy loading before printing.
+    window.addEventListener("beforeprint", (event) => {
+      $('img').each(() => {
+        $(this).removeAttr('loading'));
+      }
+    }
+
+  };
+})(Drupal);
--- a/common_design.libraries.yml
+++ b/common_design.libraries.yml
@@ -118,6 +118,8 @@ cd-other:
       libraries/cd/cd-other/cd-skip-link.css: {}
     theme:
       libraries/cd/cd-other/cd-print.css: {}
+  js:
+    js/cd-print.js: {}
 
 cd-core:
   dependencies:


js/cd-print.js
```diff
@@ -0,0 +1,21 @@
...
+((Drupal) => {
+  'use strict';
+
+  Drupal.behaviors.cdPrint = {
+    attach: function (context, settings) {
+
+      // Remove lazy loading before printing.
+      window.addEventListener('beforeprint', (event) => {
+        $('img').each(() => {
+          $(this).removeAttr('loading');
```
orakili, line 15: `$` assumes that jquery is globally available. Iframe also can be lazy [loaded](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/loading).

So better be more generic and use vanilla JS:

```suggestion
        document.querySelectorAll('[loading="lazy"]').forEach(element => {
          element.removeAttribute('loading');
        });
```

**Notes**

This may not be enough to force load the images and iframes, depending on the browsers. This needs testing and if that doesn't work, re-setting the `src` attribute for example may do the trick.

This may also require adjusting the waiting time on the snap tool.

orakili (changes requested):

js/cd-print.js
lazysoundsystem: The patch is now on the feature instance, where it seems to be working as intended: https://stage.response-reliefweb-int.ahconu.org/fr/central-african-republic/bulletin-humanitaire/reponse-durgence

js/cd-print.js
```diff
@@ -0,0 +1,21 @@
...
+((Drupal) => {
+  'use strict';
+
+  Drupal.behaviors.cdPrint = {
+    attach: function (context, settings) {
+
+      // Remove lazy loading before printing.
+      window.addEventListener('beforeprint', (event) => {
+        document.querySelectorAll('[loading="lazy"]').forEach(element => {
+          element.removeAttr('loading');
```
orakili, line 15:
```suggestion
          element.removeAttribute('loading');
```

`removeAttr` is jQuery, please use [removeAttribute](https://developer.mozilla.org/en-US/docs/Web/API/Element/removeAttribute).

orakili (changes requested):

js/cd-print.js
lazysoundsystem: ouch. sorry.

orakili (approved):
The code looks ok but not sure if that fixes the problems. As mentioned in the ticket, what was tested might not have been a good example.

I would be hesitant to add another piece of javascript to all the sites using the common design without ensuring that it actually does something useful.

lazysoundsystem:
Agreed, adding 'blocked' label until I can prove this helps.

PR #511, opened 2025-12-09
chore(feat): HID is being decommissioned and nobody should use it.

cafuego:
Refs: OPS-11879


cafuego:
Of course, in many cases we will also need a change on the custom subtheme in each site repo.

orakili (approved):

